PR #3, opened 2022-06-23
feat(pool): import client::pool from hyper

dswij:
I'm giving https://github.com/hyperium/hyper/issues/2860 a go, but I'm not sure I'm going in the right direction.

Appreciate if you could take a look when you have time @seanmonstar 

---

Some questions:

Some part of `pool` uses `hyper::Errors`

```log
error[E0624]: associated function `new_canceled` is private
   --> src/client/pool.rs:592:41
    |
592 | ...                   HyperError::new_canceled().with(CheckoutIsClosedError)
    |                                   ^^^^^^^^^^^^ private associated function
    |
   :::
    |
247 |     pub(super) fn new_canceled() -> Error {
    |     ------------------------------------- private associated function defined here
```

I'd imagine that it's probably best to expose these functions from `hyper::Error`, but I'm not sure that this would follow the plan for `hyper-util`?

### DIFF
--- a/src/client/pool.rs
+++ b/src/client/pool.rs
@@ -1,0 +1,1048 @@
+#![allow(dead_code)]
+
+use std::collections::{HashMap, HashSet, VecDeque};
+use std::error::Error;
+use std::fmt;
+use std::future::Future;
+use std::marker::Unpin;
+use std::ops::{Deref, DerefMut};
+use std::pin::Pin;
+use std::sync::{Arc, Mutex, Weak};
+use std::task::{self, Poll};
+
+#[cfg(not(feature = "runtime"))]
+use std::time::{Duration, Instant};
+#[cfg(feature = "runtime")]
+use tokio::time::{Duration, Instant, Interval};
+
+use futures_channel::oneshot;
+use tracing::{debug, trace};
+
+use super::client::Ver;
+use crate::common::{exec::Exec, ready};
+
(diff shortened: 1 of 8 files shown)


seanmonstar:
Oh right... Hm, I think this is something we'll need to experiment with, and won't have the full solution when merging this PR, and that's OK. We might just return a `hyper::Error` created with a [`Code::CANCELED`](https://github.com/hyperium/hyper/issues/2845), but maybe not. It could be that we want a `client::pool::Error` type that is separate.

It might just be easier to start with returning a `Box<dyn std::error::Error + Send + Sync>`, and we can change it later once it's been decided.

dswij:
I've replaced the `hyper::Error`s with `GenericError` for now.

src/client/pool.rs
```diff
@@ -0,0 +1,1050 @@
...
+    fn downgrade(arc: &Arc<T>) -> Self {
+        WeakOpt(Some(Arc::downgrade(arc)))
+    }
+
+    fn upgrade(&self) -> Option<Arc<T>> {
+        self.0.as_ref().and_then(Weak::upgrade)
+    }
+}
+
+#[cfg(test)]
```
seanmonstar, line 795:
```suggestion
#[cfg(all(test, not(miri)))]
```

Looks like Miri is grumpy with some of the tests in here (sigh). This should make it stop :)

seanmonstar (approved):
Thanks!

PR #4, opened 2022-07-16
feat(rt): add TokioExecutor

tomkarw:
Hey Sean,
I decided to give https://github.com/hyperium/hyper/issues/2861 a go. I'd like to ramp up the contributions after that.

Not certain what exactly is required, feel free to require additional changes.
I tried to write some basic test, but I don't know of a good way to wait for a spawned task without retaining the handle.

Also, I have a question. Why does the `Executor` trait require a generic parameter, wouldn't it make sense to only require it on `execute` function? I.e.:
```
fn execute<Fut>(&self, fut: Fut)
```

Closes https://github.com/hyperium/hyper/issues/2861

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,5 +5,6 @@ pub use crate::error::{GenericError, Result};
 
 pub mod client;
 pub mod common;
+pub mod rt;
 
 mod error;
--- a/src/rt/mod.rs
+++ b/src/rt/mod.rs
@@ -1,0 +1,4 @@
+//! Runtime utilities
+
+/// Implementation of [`hyper::rt::Executor`] that utilises [`tokio::spawn`].
+pub mod tokio_executor;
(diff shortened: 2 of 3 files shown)


davidpdrsn:
> Also, I have a question. Why does the `Executor` trait require a generic parameter, wouldn't it make sense to only require it on `execute` function? I.e.:
> 
> ```
> fn execute<Fut>(&self, fut: Fut)
> ```

Then you wouldn't be able to make an executor that only works for certain types of futures. Someone might want `impl Executor<MyFuture> for MyExecutor`

seanmonstar:
Thanks for the PR! Looks like cargo fmt has its own opinions 🙈

seanmonstar:
Looks like Miri doesn't like running Tokio in tests, You could add a `#[cfg(not(miri)]` attribute on it.

seanmonstar (approved):

PR #5, opened 2022-07-24
feat(client): move client to hyper-util

dswij:
Closes https://github.com/hyperium/hyper/issues/2860

Move client (and its parts) to hyper-util.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,8 @@ tracing = { version = "0.1", default-features = false, features = ["std"] }
 tokio = { version = "1", features = ["net", "rt", "time"] }
 tower-service = "0.3"
 tower = { version = "0.4", features = ["util"] }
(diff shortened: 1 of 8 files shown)


dswij:
Made some progress, but `client::conn` requires the use of `hyper::proto` which is private.

dswij:
@tomkarw Ah, I had been planning to do this when I have the time, but didn't get the chance to. Please do go ahead and take over if you want to. 

IIRC, there's some part of `hyper::h2` that is private and is used in `hyper::proto`. I was looking for alternative to copying this

tomkarw:
The `hyper` repo is not the one you last saw 😉. I'll see what needs to be done.

dswij:
> The `hyper` repo is not the one you last saw 😉. I'll see what needs to be done.

That's fair 😄

Issue #7, opened 2022-08-20
Usage of #[cfg(feature = ...)] in code without  the corresponding features in Cargo.toml


tomkarw:
When trying to run `cargo clippy` I encountered some warnings, but thought it was due to me not enabling all features.
That didn't work either, turns out these feature flags don't exist in `Cargo.toml`.  
Probably during port of code from hyper, somebody forgot to also create the necessary features.

If found: 
* `runtime`
* `tcp`
* `http1`
* `http2`

We should probably add the features and maybe make most imports optional.
If there are `clippy` warnings still afterwards, I'd like to fix that as well.

(As always) I'll be happy to do it once given the green light.

tomkarw:
Update is that the project doesn't compile if I add and enable the missing feature flags o.0.

`crate::common::Never` is a type that is referenced but doesn't exist.

PR #8, opened 2022-08-20
fix: add missing feature flags

tomkarw:
Closes https://github.com/hyperium/hyper-util/issues/7.

This PR:
* add missing feature flags that were referenced in code, but not in `Cargo.toml`: `runtime`, `tcp`, `http1`, `http2`
* copies `Never` struct from `hyper` and uses it in `client` module to make it compile when features are enabled
* fixes 3 Clippy warnings about preferring `Copy` to `Clone`

There is one test still failing: `client::connect::http::tests::client_happy_eyeballs`. It's a bit over my head, so I didn't attempt to fix it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ tower-service = "0.3"
 tower = { version = "0.4", features = ["util"] }
 
 [dev-dependencies]
-tokio = { version = "1", features = ["macros"] }
+tokio = { version = "1", features = ["macros", "test-util"] }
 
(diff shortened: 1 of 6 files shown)


tomkarw:
The fact that the all checks pass, even though there is a failing test, is worrying. It's also the reason code that doesn't compile was merged.

I'll make `cargo check` and `cargo test` use `--all-features` flag.

seanmonstar (approved):

PR #9, opened 2022-10-08
feat(client): `HttpConnector::tcp_keepalive_interval` and `HttpConnec…

hansonchar:
…tor::tcp_keepalive_retries` (#2991)

    If the platform supports setting the options, otherwise it's a noop.

Port from https://github.com/hyperium/hyper/pull/2991

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,10 @@ hyper = "0.14.19"
 futures-channel = "0.3"
 futures-util = { version = "0.3", default-features = false }
 http = "0.2"
+
+# Necessary to overcome msrv check of rust 1.49, as 1.15.0 failed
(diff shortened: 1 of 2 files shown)


hansonchar:
There is an `msrv` failure in the workflow that seems to be caused by `once_cell v1.15.0`.  I see the workflow succeeded 2 months ago when `once_cell v1.13.1` was used.  

Should I add the explicit dependency of once_cell:

```
once_cell = "1.13.1"
```

to overcome this failure?  Let me try...

hansonchar:
Pining `once_cell` to `1.40.0` fixes the [`MSRV 1.49` check](https://github.com/hansonchar/hyper-util/actions/runs/3220577260/jobs/5267459543).

PR #11, opened 2022-11-05
feat(server): add AutoConnection

programatik29:
Closes hyperium/hyper#2852.

This utility automatically detects HTTP protocol and uses its implementation.


src/server/conn.rs
```diff
@@ -0,0 +1,180 @@
...
+    server::conn::{http1, http2},
+    service::Service,
+};
+use std::{error::Error as StdError, io::Cursor, marker::Unpin};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite};
+
+const H2_PREFACE: &[u8] = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n";
+
+/// Http1 or Http2 connection.
+pub struct AutoConnection {
```
seanmonstar, line 15: It looks like this is a builder. I'm curious about your opinion, what do you think about `server::conn::auto::Builder`?

src/server/conn.rs
```diff
@@ -0,0 +1,180 @@
...
+    h1: http1::Builder,
+    h2: http2::Builder<TokioExecutor>,
+}
+
+impl AutoConnection {
+    /// Create a new AutoConn.
+    pub fn new() -> Self {
+        Self {
+            h1: http1::Builder::new(),
+            h2: http2::Builder::new(TokioExecutor::new()),
```
seanmonstar, line 25: Likely want to allow the user to provide an executor themselves.

src/server/conn.rs
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    /// Http1 configuration.
+    pub fn h1(&mut self) -> &mut http1::Builder {
```
seanmonstar, line 30: Perhaps we should use the longer names, `http1()` and `http2()`?

Hm, that _might_ be a case that the individual types should remove the stutter from their methods, so that `auto.http1().http1_writev(true)` were just `auto.http1().writev(true)`...

Then again, it wouldn't be possible to chain configuration. You'd almost want this kind of API:

```rust
auto
    .http1().writev(true)
    .http1().max_read_buf(32 * 1024)
    .http2().max_reset_streams(12)
    // this should use the `auto` serve, not the http2 serve
    .serve_connection(io, svc)
```

src/server/conn.rs
```diff
@@ -0,0 +1,180 @@
...
+
+                if buf[0..len] != H2_PREFACE[0..len] {
+                    break Protocol::H1;
+                }
+            } else {
+                break Protocol::H2;
+            }
+        };
+
+        let io = PrependAsyncRead::new(Cursor::new(buf), io);
```
seanmonstar, line 66: There was a similar-ish thing in `hyper::common::io::Rewind`, I don't know if it helps to compare. We can always tweak the internals to optimize, though.

seanmonstar (changes requested):
Excellent start, thanks! I appreciate how the tests are really easy to follow, that it indeed works for both versions.

Do we need to bump the CI config to make those jobs happy? Could do so in a separate PR that is easier to review/merge.

src/server/conn.rs
programatik29: I was a bit lazy there. I think `Builder` api makes more sense.

src/server/conn.rs
programatik29: That one looks shorter. We should be able to reuse it.

src/server/conn.rs
programatik29: I think going for the builder api is the way.

programatik29:
> Do we need to bump the CI config to make those jobs happy?

I have no idea what is wrong with CI. It seems like the source is dependencies.

seanmonstar:
I'm increasing the MSRV in #13, I think the Miri failures need those tests to be disabled for Miri (`#[cfg(not(miri))]`).

src/server/conn.rs
programatik29: I managed to do it by making `serve_connection` require `Executor<BoxFuture<'static, ()>>` bounds and creating an executor that does the conversion. Now the question is should we somehow make `TokioExecutor` default? If yes, how?

src/server/conn/auto.rs
```diff
@@ -0,0 +1,409 @@
...
+    http1: http1::Builder,
+    http2: http2::Builder<TokioExecutor>,
+}
+
+impl Builder {
+    /// Create a new auto connection builder.
+    pub fn new() -> Self {
+        Self {
+            http1: http1::Builder::new(),
+            http2: http2::Builder::new(TokioExecutor::new()),
```
bossmc, line 29: Could this constructor take an `http1::Builder` and `http2::Builder`, then the user can create two configurations (and set up executors/timer-handlers etc.) then the AutoConnection simply handles the protocol selection then defers to the appropriately configured HTTP stack.

As it is the AutoConnection has to expose forwarded configuration hooks for each function on the two builders and, if it misses one (e.g. because a new one is added in the hyper upstream) that configuration isn't usable until a partner change is made here.

(An alternative might be to expose `fn get_http1(&mut self) -> &mut http1::Builder` but it feels simpler to just make an HTTP1 configuration and an HTTP2 configuration then merge them together).

src/server/conn/auto.rs
```diff
@@ -0,0 +1,409 @@
...
+        let mut buf = Vec::new();
+
+        let protocol = loop {
+            if buf.len() < 24 {
+                io.read_buf(&mut buf).await?;
+
+                let len = buf.len().min(H2_PREFACE.len());
+
+                if buf[0..len] != H2_PREFACE[0..len] {
+                    break Protocol::H1;
```
bossmc, line 286: The HTTP/1 stack has an error code for "this is an HTTP/2 stream" so it's possible to just start with HTTP/1 normally and then rewind and switch to HTTP/2 if you see that error.  Uncertain which is nicer/better but just mentioning it in case it's interesting.

Something like https://github.com/hyperium/hyper/pull/3084/files#diff-4e236d9f0f3f026b7b233382e56a937697d22b5be15b6faf38a479fc722de25aR550-R597 (which I wrote before I saw this MR)

Cargo.toml
```diff
@@ -15,10 +15,12 @@ edition = "2018"
 publish = false # no accidents while in dev
 
 [dependencies]
-hyper = "=1.0.0-rc.1"
+hyper = { version = "1.0.0-rc.1", features = ["server", "http1", "http2"] }
```
bossmc, line 18: You could hide this behind a feature of `hyper-util` (so someone using some other part of `hyper-util` doesn't get forced to pick up http1, http2 and server support with all the extra dependencies that might pull in).

src/server/conn/auto.rs
```diff
@@ -0,0 +1,409 @@
...
+        self
+    }
+    /// Set the maximum buffer size for the connection.
+    ///
+    /// Default is ~400kb.
+    ///
+    /// # Panics
+    ///
+    /// The minimum value allowed is 8192. This method panics if the passed `max` is less than the minimum.
+    #[cfg(feature = "http1")]
```
bossmc, line 109: You're a bit inconsistent with `feature` usage across this change, the AutoConnection probably can assume both are already present since it has to hold an `http1::Builder` and `http2::Builder` so these checks are probably not needed.

bossmc (commented):
Hi, @seanmonstar pointed me at this MR since I'd implemented something similar as part of https://github.com/hyperium/hyper/pull/3084, I've added some comments from comparing how I did it in hyper directly compared to how you've done things here.

src/server/conn/auto.rs
programatik29: True. Side effects of copy pasting.

src/server/conn/auto.rs
programatik29: Current approach makes more sense IMO.

src/server/conn/auto.rs
programatik29: Maybe combine both approaches somehow? Have a nice builder API and allow passing http1 and http2 builders.

src/server/conn/auto.rs
bossmc: Thinking about it - I agree (and, because this is a "serve_connection" call, it fits into my FFI usage just fine 👍)

src/common/rewind.rs
```diff
@@ -0,0 +1,161 @@
...
+        cx: &mut task::Context<'_>,
+        buf: &mut ReadBuf<'_>,
+    ) -> Poll<io::Result<()>> {
+        if let Some(mut prefix) = self.pre.take() {
+            // If there are no remaining bytes, let the bytes get dropped.
+            if !prefix.is_empty() {
+                let copy_len = cmp::min(prefix.len(), buf.remaining());
+                // TODO: There should be a way to do following two lines cleaner...
+                buf.put_slice(&prefix[..copy_len]);
+                prefix.advance(copy_len);
```
jwarlander, line 66: Wouldn't this work?
```suggestion
                buf.put_slice(&prefix.split_to(copy_len));
```

programatik29: I didn't write this code but current one looks more efficient.

programatik29: @seanmonstar

jwarlander: > I didn't write this code but current one looks more efficient.

Ah, sorry - thought it was your TODO note 😅 But yes, the original approach is most likely better performing as it's just a slice operation for `&[u8] `, though it's unclear by how much, as the documentation for `split_at` indicates it should be reasonably fast too.

src/server/conn.rs
seanmonstar: Now with RC3 out, you should be able to use the `hyper::rt::bounds` that you added, right?

src/common/rewind.rs
seanmonstar: Fixing it or leaving it alone is both fine.

src/server/conn.rs
programatik29: Yep.

seanmonstar:
What's the status here? I see you added some new commits, is this ready for a re-review?

programatik29:
> What's the status here? I see you added some new commits, is this ready for a re-review?

Yes.

src/server/conn/auto.rs
```diff
@@ -0,0 +1,503 @@
...
+}
+
+/// Http2 part of builder.
+pub struct Http2Builder<'a, E> {
+    inner: &'a mut Builder<E>,
+}
+
+impl<E> Http2Builder<'_, E> {
+    /// Http1 configuration.
+    pub fn http1(&mut self) -> Http1Builder<'_, E> {
```
seanmonstar, line 246: Oh, now I see how you've done this. That's clever! I like the way it looks down in the unit tests.

I was slightly confused why both sub-builders had `serve_connection` methods, and now seeing this one here, I understand. I have a suggestion, and while I kinda like it, I'm open to hearing why it might be _worse_:

What if each method on the sub-builder returned, not `&mut Self`, but `&mut Builder<E>`? The builders in Finagle work the same way: a sub-builder is only usable once, and then you get back the full thing. It does make it easier to access options that affect *both* versions, if we were to add any.

So:

```rust
Builder::new(exec)
    .http1().keep_alive(true)
    .http1().writev(true)
    .http2().enable_connect_protocol(true)
    .max_send_buf(16324) // hypothetical, applies to both versions
    .serve_connection(io, svc);
```

seanmonstar (commented):
This is excellent, I love it! I had one cosmetic idea that I wrote up inline.

src/server/conn/auto.rs
programatik29: I'm open to it. I think we should ask users somehow on which one they prefer. Personally, I can work with both.

seanmonstar (approved):
There was an informal poll, and the current API was preferred, so let's merge with that. Thanks for all the work putting this together!

programatik29:
@seanmonstar `client` module is generating errors which isn't related to this PR.

seanmonstar:
@programatik29 fixed.

PR #12, opened 2022-11-30
implement generic key for no feature flag

EItanya:
This PR attempts to make the client pooling `Key` generic so that the caller can decide how they wish to pool the client connections.

As of writing, it does not work with the `tokio` runtime, I am relatively new to rust and was having some issues getting on of the futures functions to play nicely, would love so help.

### DIFF
--- a/src/client/pool.rs
+++ b/src/client/pool.rs
@@ -2,8 +2,9 @@
 
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::error::Error;
-use std::fmt;
+use std::fmt::{self, Debug};
 use std::future::Future;
+use std::hash::Hash;
 use std::marker::Unpin;
 use std::ops::{Deref, DerefMut};
 use std::pin::Pin;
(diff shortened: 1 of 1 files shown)


src/client/pool.rs
```diff
@@ -42,6 +43,8 @@ pub(super) trait Poolable: Unpin + Send + Sized + 'static {
     fn can_share(&self) -> bool;
 }
 
+pub trait Key: Eq + Hash + Clone + Debug + Unpin {}
```
seanmonstar, line 46: Do we need the separate trait?

It looks like this may exist just to reduce the list at `K: Eq + Hash + ...`. But if the trait doesn't provide anything new, it would provide a hurdle to users: they'd need to `impl Key for MyThing`. That'd be more annoying if `MyThing` were from a different crate, so they _couldn't_ implement it.

seanmonstar (commented):
Looks pretty on point, just one thought inline.

src/client/pool.rs
EItanya: That's na great point, new to `Rust` things!!

howardjohn: @seanmonstar we could do a blanket implementation? Or do you prefer putting all the individual traits on each bounds?

PR #15, opened 2022-12-15
pool: make relevant items `pub`

howardjohn:
Currently `pool` is not pub, and nothing `pub` uses it, so it isn't accessible. This exposes them for use.

Discussed here: https://discordapp.com/channels/500028886025895936/670880858630258689/1053021031805227109

### DIFF
--- a/src/client/pool.rs
+++ b/src/client/pool.rs
@@ -24,7 +24,7 @@ use crate::common::{exec::Exec, ready};
 
 // FIXME: allow() required due to `impl Trait` leaking types to this lint
 #[allow(missing_debug_implementations)]
-pub(super) struct Pool<T, K: Key> {
+pub struct Pool<T, K: Key> {
     // If the pool is disabled, this is None.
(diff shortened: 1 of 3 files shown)


src/client/pool.rs
```diff
@@ -33,7 +33,7 @@ pub(super) struct Pool<T> {
 // This is a trait to allow the `client::pool::tests` to work for `i32`.
 //
 // See https://github.com/hyperium/hyper/issues/1429
-pub(super) trait Poolable: Unpin + Send + Sized + 'static {
+pub trait Poolable: Unpin + Send + Sized + 'static {
```
seanmonstar, line 36: I'm fine with just making it all public to start with, to unblock people, but I do think the exposed design could be improved, by making the `Pool` basically a `MakeService` over another `MakeService` (a caching layer over the `Connect` stuff).

seanmonstar (approved):

PR #21, opened 2023-02-18
Fix legacy client feature config

chrislearn:
Legacy Client's feature config seems not correct, and example has compile error.


seanmonstar:
Thank you! Looks like rustfmt is grumpy.

seanmonstar (approved):
Wonderful, thank you! :)

PR #22, opened 2023-02-22
Unlock once_cell version

chrislearn:
In CI MSRV is set to 1.56. It seems lock once_cell to 1.14 is not necessary，and this lock cause conflict with many other crates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ futures-channel = "0.3"
 futures-util = { version = "0.3", default-features = false }
 http = "0.2"
 
-once_cell = "1.17"
+once_cell = "1.14"
 
 pin-project-lite = "0.2.4"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -20,8 +20,7 @@ futures-channel = "0.3"
 futures-util = { version = "0.3", default-features = false }
 http = "0.2"
 
-# Necessary to overcome msrv check of rust 1.49, as 1.15.0 failed
-once_cell = "=1.14"
+once_cell = "1.17"
```
seanmonstar, line 23: Do we _require_ `.17`? What's the minimum version we need? That should be what is set here. Then, anyone else who is holding back upgrades isn't forced by us.

seanmonstar (approved):

PR #25, opened 2023-03-09
client: add timer() option

howardjohn:
Currently, a timer must be provided to use http2. However, this has no option to actually configure this. This PR introduces a new function to set the timer. With this, I am able to successfully make HTTP2 calls with the Client (as long as there is no error - currently we hit a todo!() on errors)

### DIFF
--- a/src/client/legacy.rs
+++ b/src/client/legacy.rs
@@ -14,6 +14,7 @@ use std::time::Duration;
 use futures_util::future::{self, Either, FutureExt, TryFutureExt};
 use http::uri::Scheme;
 use hyper::header::{HeaderValue, HOST};
+use hyper::rt::Timer;
 use hyper::{body::Body, Method, Request, Response, Uri, Version};
 use tracing::{debug, trace, warn};
 
@@ -1346,6 +1347,22 @@ impl Builder {
     #[cfg_attr(docsrs, doc(cfg(feature = "http2")))]
     pub fn http2_max_concurrent_reset_streams(&mut self, max: usize) -> &mut Self {
         self.h2_builder.max_concurrent_reset_streams(max);
+        self
+    }
+
+    /// Provide a timer to execute background HTTP2 tasks
+    ///
+    /// See the documentation of [`h2::client::Builder::timer`] for more
+    /// details.
+    ///
+    /// [`h2::client::Builder::timer`]: https://docs.rs/h2/client/struct.Builder.html#method.timer
+    pub fn timer<M>(&mut self, timer: M) -> &mut Self
+    where
+        M: Timer + Send + Sync + 'static,
+    {
+        #[cfg(feature = "http2")]
+        self.h2_builder.timer(timer);
+        // TODO(https://github.com/hyperium/hyper/issues/3167) set for pool as well
         self
     }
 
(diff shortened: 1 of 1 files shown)


src/client/legacy.rs
```diff
@@ -1347,6 +1348,22 @@ impl Builder {
...
 
+    /// Provide a timer to execute background HTTP2 tasks
+    ///
+    /// See the documentation of [`h2::client::Builder::timer`] for more
+    /// details.
+    ///
+    /// [`h2::client::Builder::timer`]: https://docs.rs/h2/client/struct.Builder.html#method.timer
+    #[cfg(feature = "http2")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "http2")))]
+    pub fn http2_timer<M>(&mut self, timer: M) -> &mut Self
```
seanmonstar, line 1359: Doesn't the pool need a timer too? Should we just call this method `timer()`?

seanmonstar (commented):
What do you mean it fails without a timer? That should be fixed if that's the case, a timer shouldn't be _required_.

howardjohn:
> What do you mean it fails without a timer? That should be fixed if that's the case, a timer shouldn't be _required_.

Right now I am getting a panic here https://github.com/hyperium/hyper/blob/9ed175d1545fa338b6212c4f48be9b3d12821c7a/src/common/time.rs#L61 called from Client::send_request, which eventually calls `__timer.sleep(interval)` as part of the keepalive for `channel`.

I am not seeing where pool has a timer, looks like spawn_idle_interval alwys uses tokio::time::interval if I am reading things right

seanmonstar:
Ah, is that because you're setting some of the HTTP/2 keep-alive options?

src/client/legacy.rs
howardjohn: > Ah, is that because you're setting some of the HTTP/2 keep-alive options?

Ah... that is it. I didn't make that connection.

I updated the name to `timer()` and opened https://github.com/hyperium/hyper/issues/3167 to track making Pool use it as well. I am hoping to pick up that task as well but won't be able to get to it for a few days at least so would prefer it separate

seanmonstar: > I updated the name to timer()

Did you mean to push a new commit? 😅

howardjohn: Oops, should be pushed now. Thanks!

src/client/legacy.rs
```diff
@@ -1347,6 +1348,22 @@ impl Builder {
...
+    /// See the documentation of [`h2::client::Builder::timer`] for more
+    /// details.
+    ///
+    /// [`h2::client::Builder::timer`]: https://docs.rs/h2/client/struct.Builder.html#method.timer
+    pub fn timer<M>(&mut self, timer: M) -> &mut Self
+    where
+        M: Timer + Send + Sync + 'static,
+    {
+        #[cfg(feature = "http2")]
+        self.h2_builder.timer(timer);
```
howardjohn, line 1362: This does nothing without `http2` feature, but I think that is ok since it will once we add it to the pool(?)

seanmonstar: Yes, that sounds right.

howardjohn:
Hey @seanmonstar , just coming back to this. I think this is ready to go from my POV, let me know what you think when you have a chance?

seanmonstar (approved):

PR #33, opened 2023-09-07
Update socket2 dependency in Cargo.toml.

m-hilgendorf:
The current tip of main does not compile on Linux due to an invalid version of socket2.


seanmonstar:
Thank you for the PR, I had to update both the MSRV and this dependency at the same time, which I did in #34.

PR #35, opened 2023-09-15
chore: use futures_util::ready

tottoto:
`hyper-util` could use `futures_util::ready`.


davidpdrsn:
Or even https://doc.rust-lang.org/stable/std/future/fn.ready.html

tottoto:
I think the stabilized API is https://doc.rust-lang.org/stable/std/task/macro.ready.html, which requires Rust 1.64.

jplatte:
Typo: `utill`

seanmonstar (approved):

PR #38, opened 2023-09-28
Added `.with_upgrades()` to connection

rakshith-ravi:
In all honesty, I really don't know if it's possible to have this happen in the same function with a `?Send` way, but I couldn't get that to work. So I added a new function that requires the IO stream to be `Send`, which automatically upgrades HTTP/1 connections.

Also, does HTTP/2 not require upgrades? I was looking for a similar `.with_upgrades()` function in `self.http2.serve_connection`, and when I couldn't find it, I tried adding it, and realized that there was no `upgrades` module at all. Is this inherently present in HTTP/2 or is this just something that we haven't implemented yet?

This is required by tokio-rs/axum#2241

### DIFF
--- a/src/server/conn/auto.rs
+++ b/src/server/conn/auto.rs
@@ -86,6 +86,35 @@ impl<E> Builder<E> {
 
         Ok(())
     }
+
+    /// Bind a connection together with a [`Service`], with the ability to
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):

rakshith-ravi:
Woohoo!!!! Thanks a lot!

PR #40, opened 2023-10-15
Drop Send requirements on the future returned by the service

LukeMathWalker:
The current bounds appear to be too strict when it comes to `Send`ness, preventing the usage of `auto` in combination with an implementation of `Executor` that's designed to spawn `!Send` futures.


seanmonstar (approved):
Thanks!

PR #46, opened 2023-11-10
feat(service): add adapter for tower services

davidpdrsn:
This adds

- `TowerToHyperService` which converts a tower service into a hyper service.
- `Builder::serve_connection_tower` and `Builder::serve_connection_with_upgrades_tower` for serving a tower service directly. The adapter will be applied internally.
- An example showing how to use it since there was previously only a client example.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,5 +8,10 @@ pub mod client;
 mod common;
 pub mod rt;
 pub mod server;
+#[cfg(all(
+    any(feature = "http1", feature = "http2"),
+    any(feature = "server", feature = "client")
+))]
+pub mod service;
 
 mod error;
--- a/src/server/conn/auto.rs
+++ b/src/server/conn/auto.rs
@@ -87,6 +87,40 @@ impl<E> Builder<E> {
         Ok(())
     }
 
+    /// Bind a connection together with a [`tower_service::Service`].
+    #[cfg(all(
+        any(feature = "http1", feature = "http2"),
+        any(feature = "server", feature = "client")
+    ))]
+    pub async fn serve_connection_tower<I, S, B>(&self, io: I, service: S) -> Result<()>
+    where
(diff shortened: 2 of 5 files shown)


src/server/conn/auto.rs
```diff
@@ -87,6 +88,36 @@ impl<E> Builder<E> {
...
+                    .await?
+            }
+            Version::H2 => {
+                self.http2
+                    .serve_connection(io, TowerToHyperService::new(service))
+                    .await?
+            }
+        }
+
+        Ok(())
```
davidpdrsn, line 118: Ideally we'd just call `serve_connection` but that requires `S: 'static`. Not totally sure why that is. Copy-pasting the impl doesn't require that 🤔

src/lib.rs
```diff
@@ -8,5 +8,10 @@ pub mod client;
 mod common;
 pub mod rt;
 pub mod server;
+#[cfg(all(
+    any(feature = "http1", feature = "http2"),
+    any(feature = "server", feature = "client")
+))]
```
davidpdrsn, line 14: Perhaps hyper should have a `service` feature so we don't need to do stuff like this.

src/server/conn/auto.rs
```diff
@@ -87,6 +87,40 @@ impl<E> Builder<E> {
         Ok(())
     }
 
+    /// Bind a connection together with a [`tower_service::Service`].
+    #[cfg(all(
+        any(feature = "http1", feature = "http2"),
+        any(feature = "server", feature = "client")
+    ))]
+    pub async fn serve_connection_tower<I, S, B>(&self, io: I, service: S) -> Result<()>
```
davidpdrsn, line 95: Been thinking that axum also needs a version that uses `MakeService` for https://docs.rs/axum/latest/axum/struct.Router.html#method.into_make_service_with_connect_info

davidpdrsn:
I wanna do some more thinking about how to best do this. So I'll close this PR for now and submit PRs with smaller pieces.

PR #49, opened 2023-11-16
feat: moved connect module to legacy

kaffarell:
Moved the `connect` module to the legacy namespace to make space for new changes. Renamed legacy.rs -> client.rs and also moved to that namespace.

Fixes: hyperium/hyper#3414


examples/client.rs
```diff
@@ -2,7 +2,7 @@ use std::env;
 
 use http_body_util::Empty;
 use hyper::Request;
-use hyper_util::client::{connect::HttpConnector, legacy::Client};
+use hyper_util::client::legacy::{client::Client, connect::HttpConnector};
```
seanmonstar, line 5: To mirror what was "legacy" (0.14), I think we could keep the import to `client::legacy::{Client, connect::HttpConnector}`.

kaffarell: Thanks for the review! Fixed it right now.

seanmonstar:
Looks like the feature flags need to be adjusted.

PR #50, opened 2023-11-16
clean up exposed crate features

seanmonstar:
- Put `TokioExecutor` and `TokioIo` into a single `rt::tokio` module.
- The `rt::tokio` module is behind a `tokio` crate feature.
- Changed `auto` crate feature to `server-auto`.
- Added `client-legacy` crate feature, with `client::legacy` behind it.
- Removed `tcp` and `runtime` features, code now depends on the `tokio` feature.


seanmonstar:
@howardjohn I've adjusted some of the features needed to use the legacy client, could you confirm for me that with the changes, your use case works?

howardjohn:
> @howardjohn I've adjusted some of the features needed to use the legacy client, could you confirm for me that with the changes, your use case works?

Thanks! This works for our case, but looks like #49 broke us. With https://github.com/hyperium/hyper-util/pull/51 + this PR everything is working fine

PR #57, opened 2023-11-17
Feature flags: `server-auto` should imply `server`

LukeMathWalker:
As it currently stands, enabling `server-auto` doesn't actually give you access to any new item _unless_ you also enable the `server` feature.
Considering how the code is structured, I think that the `server-auto` feature flag should imply the `server` feature flag.


seanmonstar (approved):
Yep, that's how it should be. Just like client-legacy enables client. I wrote the wrong thing. Thanks!

PR #63, opened 2023-11-22
Fix `cargo minimal-versions check`

Finomnis:
In a `Cargo.toml`, the dependency versions mean that those versions are the minimum versions required for the project to build.
Sadly, the minimum versions that are given right now will cause compiler errors.

This can be checked through the tool [`cargo-minimal-versions`](https://crates.io/crates/cargo-minimal-versions).

This change modifies the minimal versions so that `cargo minimal-versions check` now succeeds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 hyper = "1.0.0"
 futures-channel = "0.3"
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Thanks!

It'd probably be a good idea to add such a check to CI, similar to [hyper has](https://github.com/hyperium/hyper/blob/210bfaa711b5da1f6756582a2e4bc3e229924800/.github/workflows/CI.yml#L277).

PR #64, opened 2023-11-23
feat: add `TowerToHyperService`

davidpdrsn:
Extracted from https://github.com/hyperium/hyper-util/pull/46. This adds `TowerToHyperService` which converts a tower service into a hyper service. Among other things, this is useful for running axum with hyper.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,5 +12,10 @@ mod common;
 pub mod rt;
 #[cfg(feature = "server")]
 pub mod server;
+#[cfg(all(
+    any(feature = "http1", feature = "http2"),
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):
Looks fine to me!

Any reason we shouldn't make it part of an optional `tower` feature, like how `tokio` enables those RT types?

jplatte:
Seems like tower is an unconditional dependency already. But maybe not public?

TannerRogalsky:
Once this is merged I'll update the [turmoil example](https://github.com/tokio-rs/turmoil/blob/main/examples/axum/src/main.rs).

PR #70, opened 2023-12-02
Added Body wrapper

WinLinux1028:
This wrapper is useful when there is the possibility of returning various types of Body.
Like this:
```Rust
async fn hello(req: Request<hyper::body::Incoming>) -> Result<Response<Body>, Infallible> {
    if req.uri().path() == "/" {
        Ok(Response::new(Body::new(Full::new(Bytes::from(
            "Hello, World!",
        )))))
    } else {
        Ok(Response::new(Body::new(Empty::<Bytes>::new())))
    }
}
```

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -1,0 +1,112 @@
+use std::pin::pin;
+
+use http::{Request, Response};
+
+/// Wrapping [http_body::Body]
+pub struct Body(
+    Box<
+        dyn http_body::Body<
+                Data = Box<dyn bytes::Buf + Send>,
+                Error = Box<dyn std::error::Error + Send + Sync>,
+            > + Unpin
+            + Send,
+    >,
+);
+
+impl Body {
+    /// Create new instance from any `Body`
+    pub fn new<B, D, E>(body: B) -> Self
+    where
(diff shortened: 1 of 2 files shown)


davidpdrsn:
This already exists in http-body-util as [BoxBody](https://docs.rs/http-body-util/0.1.0/http_body_util/combinators/struct.BoxBody.html) and [UnsyncBoxBody](https://docs.rs/http-body-util/0.1.0/http_body_util/combinators/struct.UnsyncBoxBody.html).

WinLinux1028:
No, those aren't wrap Data and Error, but this wraps.

davidpdrsn:
You generally don’t need to wrap Data. It’ll be Bytes is most cases.

For error you can use `.map_err` to box it before wrapping it in BoxBody.

WinLinux1028:
That method can certainly be used, but it is a bit verbose (especially when wrapped in Request or Response).
```Rust
let (parts, body) = response.into_parts();
let body = body.map_err(|e| Box::new(e));
let response = Response::from_parts(parts, body);
```
```Rust
let response = Body::convert_response(response);
```

seanmonstar:
As mentioned, something very similar already exists in http-body-util.

PR #72, opened 2023-12-07
feat: make builder clonable

neoeinstein:
Being able to create a single builder, which then can be cloned to enable reuse would be ideal. Also implements `Debug` for the builder.


seanmonstar (approved):
Thanks!

PR #73, opened 2023-12-08
feat: Add TokioTimer to rt::tokio module

daxhuiberts:
In the latest master branch of `hyper` the `hello` server example has been changed to include a `TokioTimer`.

This exists in `hyper`'s `support` module, but does not yet exist in `hyper-util`.

This PR adds that functionality by copying over the implementation from `hyper`.

Similar to `TokioExecutor` in this repo. It can only be created through the `new` constructor. This differs from `hyper`'s `support` module, where both `TokioExecutor` and `TokioTimer` are directly creatable as an empty struct.

### DIFF
--- a/src/rt/tokio.rs
+++ b/src/rt/tokio.rs
@@ -4,9 +4,10 @@ use std::{
     future::Future,
     pin::Pin,
     task::{Context, Poll},
+    time::{Duration, Instant},
 };
 
-use hyper::rt::Executor;
+use hyper::rt::{Executor, Sleep, Timer};
 use pin_project_lite::pin_project;
 
 /// Future executor that utilises `tokio` threads.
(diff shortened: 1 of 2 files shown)


src/rt/tokio.rs
```diff
@@ -190,6 +206,51 @@ where
...
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        self.project().inner.poll(cx)
+    }
+}
+
+impl Sleep for TokioSleep {}
+
+impl TokioSleep {
+    pub fn reset(self: Pin<&mut Self>, deadline: Instant) {
```
seanmonstar, line 249:
```suggestion
    fn reset(self: Pin<&mut Self>, deadline: Instant) {
```

seanmonstar:
Thanks for noticing and submitting this, I appreciate it a ton! ❤️

src/rt/tokio.rs
daxhuiberts: Good catch! FYI it's also `pub fn` in `hyper`'s `support::tokiort` module.

seanmonstar (approved):

loyd:
Hi, when will it be published?

PR #76, opened 2023-12-09
fix(http): divide by zero error when dividing duration by scalar

gngpp:
from hyper 0.14.x
commit：https://github.com/hyperium/hyper/pull/3355


seanmonstar (approved):
Thanks for porting this over!

PR #79, opened 2023-12-16
feat: tower should be an optional dependency

sigoden:
The `tower/tower-service` is only used in the `client/client-legacy` feature. If one only needs `server/server-auto` feature, there is no need to include them.

Testing in sigoden/dufs#321,  with this pr, there is a reduction of 0.3 M in the stripped release binary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,8 +27,8 @@ pin-project-lite = "0.2.4"
 socket2 = { version = "0.5", optional = true, features = ["all"] }
 tracing = { version = "0.1", default-features = false, features = ["std"] }
 tokio = { version = "1", optional = true, features = ["net", "rt", "time"] }
-tower-service = "0.3"
-tower = { version = "0.4.1", features = ["make", "util"] }
+tower-service ={ version = "0.3", optional = true }
+tower = { version = "0.4.1", optional = true, features = ["make", "util"] }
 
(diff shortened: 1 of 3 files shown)


tottoto:
Sounds reasonable to me.

seanmonstar (approved):
Me too, thanks!

seanmonstar:
Ah, looks like there's some `cfg` adjustments needed.

davidpdrsn:
Is this a breaking change?

tottoto:
As `TowerToHyperService` has not been released, I do not think this is a breaking change.

sigoden:
Need to adjust the cfg! of hyper.

https://github.com/hyperium/hyper-util/blob/99409f5c4059633b7e2fa8b9c2e6c110b0f2f64b/src/service.rs#L30

```rust
#[cfg(all(
    any(feature = "http1", feature = "http2"),
    any(feature = "server", feature = "client")
))]
pub use self::service::Service;
```

https://github.com/hyperium/hyper/blob/d9c5d3b9b3921e05040ef50d4aca080f563aac36/src/service/mod.rs#L30-L33

sigoden:
> As `TowerToHyperService` has not been released, I do not think this is a breaking change.

As @tottoto said, this is not a break change. @davidpdrsn

PR #80, opened 2023-12-19
feat: allow enabling http1/http2 individually

SergioBenitez:
As in the title. Previously we pulled in just the deps for http1 or just the deps for http2. This reenables that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,14 +19,14 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 hyper = "1.1.0"
-futures-channel = "0.3"
 futures-util = { version = "0.3.16", default-features = false }
 http = "1.0"
 http-body = "1.0.0"
 bytes = "1"
 pin-project-lite = "0.2.4"
+futures-channel = { version = "0.3", optional = true }
 socket2 = { version = "0.5", optional = true, features = ["all"] }
-tracing = { version = "0.1", default-features = false, features = ["std"] }
+tracing = { version = "0.1", default-features = false, features = ["std"], optional = true }
 tokio = { version = "1", optional = true, features = ["net", "rt", "time"] }
 tower-service ={ version = "0.3", optional = true }
 tower = { version = "0.4.1", optional = true, features = ["make", "util"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -59,7 +59,6 @@ client = ["hyper/client", "dep:tower", "dep:tower-service"]
 client-legacy = ["client"]
 
 server = ["hyper/server"]
-server-auto = ["server", "http1", "http2"]
```
seanmonstar, line 62: Removing this would be breaking.

seanmonstar (commented):
This could be a way to address https://github.com/hyperium/hyper/issues/3493, if we leave the `server-auto` feature in. Basically, if you have enabled `server` and `http1`, then the auto builder can exist with only the HTTP/1 parts.

What do you think @davidpdrsn?

SergioBenitez:
It seems that `auto.rs` has received some rather major changes. Happy to rebase and fix the fallout as soon as this is generally approved.

SergioBenitez:
Rebased on master.

ikrivosheev:
@seanmonstar @SergioBenitez any updates?)

SergioBenitez:
@ikrivosheev There's nothing to do on my end.

SergioBenitez:
@seanmonstar What's preventing this from getting in? I really don't want to keep rebasing as changes are made upstream.

src/server/conn/auto.rs
```diff
@@ -52,17 +81,23 @@ impl<E> Builder<E> {
...
             http1: http1::Builder::new(),
+            #[cfg(feature = "http2")]
             http2: http2::Builder::new(executor),
+            #[cfg(not(feature = "http2"))]
+            _executor: executor,
         }
     }
 
     /// Http1 configuration.
+    #[cfg(feature = "http1")]
```
seanmonstar, line 94: I think the main blocker is that this would be breaking, no? Currently, if you forgot to enable the `http1` or `http2` features, after this change, your code would no longer compile.

SergioBenitez: The only way to have this be publicly exposed before is by enabling `server-auto`. That feature now enables `http1` and `http2`, making this backwards compatible.

seanmonstar: Oh right, sorry. I just realized reading back I understood that in my previous comment.

OK, so with this PR, if you only enable `server` and `http1`, the auto builder can exist, with only the `Http1Builder` part of it. Does that sound right?

SergioBenitez: That's right.

seanmonstar (approved):

PR #82, opened 2023-12-21
Document the correct default value

psibi:
Load testing on our web application caused issues and it looks like it is because of this.

The default value is 200 as you can see from the default instantiation here: https://github.com/hyperium/hyper/blob/1d4ff3597b8e76818c8553dbfa4234cf4208c958/src/proto/h2/server.rs#L66


seanmonstar (approved):
Ah, a place we forgot to update when setting a default. Thanks for fixing it!

PR #77, opened 2023-12-09
feat: port legacy client integration tests

dswij:
part of https://github.com/hyperium/hyper/issues/2942

PR #83, opened 2023-12-22
fix(client): no keep-alive always returning an error

dswij:
fixes https://github.com/hyperium/hyper/issues/3499

This PR fixes the `hyper_util::client::legacy::Client` always returning an error without sending any request.

Found while working on https://github.com/hyperium/hyper-util/pull/77, test for this will be included in the PR.

---

After the change, the code in the issue returns
```rust
Ok(
    Response {
        status: 200,
        version: HTTP/1.0,
        headers: {
            "server": "SimpleHTTP/0.6 Python/3.11.4",
            "date": "Fri, 22 Dec 2023 13:24:20 GMT",
            "content-type": "text/html; charset=utf-8",
            "content-length": "2204",
        },
        body: Body(
            Streaming,
        ),
    },
)
```

instead of `PoolDisabled` error

### DIFF
--- a/src/client/legacy/client.rs
+++ b/src/client/legacy/client.rs
@@ -382,6 +382,14 @@ where
         &self,
         pool_key: PoolKey,
     ) -> Result<pool::Pooled<PoolClient<B>, PoolKey>, ClientConnectError> {
+        // Return a single connection if pooling is not enabled
+        if !self.pool.is_enabled() {
+            return self
+                .connect_to(pool_key)
+                .await
+                .map_err(ClientConnectError::Normal);
+        }
+
         // This actually races 2 different futures to try to get a ready
         // connection the fastest, and to reduce connection churn.
         //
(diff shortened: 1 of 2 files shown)


seanmonstar:
Hm, there's this code in the repo already:

```rust
            Either::Left((Err(err), connecting)) => {
                if err.is_canceled() {
                    connecting.await.map_err(ClientConnectError::Normal)
                } else {
                    Err(ClientConnectError::Normal(e!(Connect, err)))
                }
            }
```

Is the problem that the pool disabled error no longer results in `err.is_canceled()` being true?

dswij:
> Is the problem that the pool disabled error no longer results in err.is_canceled() being true?

Yes.

But doing it this way is kinda confusing. Explicitly selecting one future in this case instead of saying it's canceled is clearer in my opinion.

seanmonstar:
Agreed, this is clearer. As long no extra logic that should apply in both cases happens down in the select.

seanmonstar (approved):

PR #84, opened 2023-12-28
feat: add timer support for `legacy::Pool`

dswij:
Addresses https://github.com/hyperium/hyper/issues/3167

Note that previously we're using tokio's `Interval` rather than `Sleep`. This PR tries to emulate the behavior for `Pool` using `Sleep` instead, so that:

1. No extra trait for `Interval`
2. User only needs to supply a `Timer` rather than a `Timer` and also an `Interval`.

### DIFF
--- a/src/client/legacy/client.rs
+++ b/src/client/legacy/client.rs
@@ -22,7 +22,8 @@ use tracing::{debug, trace, warn};
 use super::connect::HttpConnector;
 use super::connect::{Alpn, Connect, Connected, Connection};
 use super::pool::{self, Ver};
-use crate::common::{lazy as hyper_lazy, Exec, Lazy, SyncWrapper};
+
+use crate::common::{lazy as hyper_lazy, timer, Exec, Lazy, SyncWrapper};
 
 type BoxSendFuture = Pin<Box<dyn Future<Output = ()> + Send>>;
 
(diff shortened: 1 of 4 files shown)


src/client/legacy/client.rs
```diff
@@ -1374,11 +1397,11 @@ impl Builder {
     /// [`h2::client::Builder::timer`]: https://docs.rs/h2/client/struct.Builder.html#method.timer
     pub fn timer<M>(&mut self, timer: M) -> &mut Self
     where
-        M: Timer + Send + Sync + 'static,
+        M: Timer + Clone + Send + Sync + 'static,
```
seanmonstar, line 1400: Woops. Adding this would technically be breaking.

I suppose we could just wrap it in the local `Timer` struct, and pass _that_? It will mean double Arc-dyn, which is unfortunate. Maybe we can find a clever way to indicate to `hyper::rt::Timer` that we don't need to double Arc it?

dswij: Whoops yeah you're right, forgot about this one.

For avoiding double Arc-dyn, we can probably come up with something using `Deref`  and or `TryFrom`, though I feel it's a bit hacky. (something like, if it's from local `Timer` then don't need to wrap Arc)

I was playing around with `unwrap_or_clone` just to realize it's still experimental. :sweat_smile: https://doc.rust-lang.org/std/sync/struct.Arc.html#method.unwrap_or_clone

seanmonstar (approved):

PR #85, opened 2023-12-31
Remove an unnecessary Unpin bound

sfackler:


sfackler:
Seems like this is required by h2?

PR #86, opened 2024-01-08
2862: GracefulShutdown utility

GlenDC:
Closes [#2862](https://github.com/hyperium/hyper/issues/2862)

Adds a `GracefulShutdown` utility which:

- is runtime agnostic
- can work with any future
- respects the API as requested in the feature request issue (#2862), taken also into account comments and ideas from this PR;

### DIFF
--- a/src/server/graceful.rs
+++ b/src/server/graceful.rs
@@ -3,27 +3,22 @@
 //! This module provides a [`GracefulShutdown`] type,
 //! which can be used to gracefully shutdown a server.
 //!
-//! # Example
-//!
-//! TODO
+//! See <https://github.com/hyperium/hyper-util/blob/master/examples/server_graceful.rs>
+//! for an example of how to use this.
 
 use futures_util::FutureExt;
 use pin_project_lite::pin_project;
 use slab::Slab;
 use std::{
-    fmt::{self, Debug, Display},
+    fmt::{self, Debug},
     future::Future,
     pin::{pin, Pin},
-    sync::{
-        atomic::{AtomicBool, AtomicUsize},
-        Arc, Mutex,
-    },
+    sync::{atomic::AtomicUsize, Arc, Mutex},
     task::{self, Poll, Waker},
 };
 use tokio::select;
 
 /// A graceful shutdown watcher
-#[derive(Clone)]
 pub struct GracefulShutdown {
     state: Arc<GracefulState>,
 }
@@ -46,31 +41,51 @@ impl GracefulShutdown {
         Self {
             state: Arc::new(GracefulState {
                 counter: AtomicUsize::new(0),
-                shutdown: AtomicBool::new(false),
                 waker_list: Mutex::new(Slab::new()),
             }),
         }
     }
 
+    /// Get a graceful shutdown watcher
+    pub fn watcher(&self) -> GracefulWatcher {
+        self.state.subscribe();
+        GracefulWatcher {
+            state: self.state.clone(),
+        }
+    }
+
+    /// Wait for a graceful shutdown
+    pub fn shutdown(self) -> GracefulWaiter {
+        GracefulWaiter {
+            state: GracefulWaiterState {
+                state: self.state.clone(),
+                key: None,
+            },
+        }
+    }
+}
+
+/// A graceful shutdown watcher.
+pub struct GracefulWatcher {
+    state: Arc<GracefulState>,
+}
+
+impl Drop for GracefulWatcher {
+    fn drop(&mut self) {
+        self.state.unsubscribe();
+    }
+}
+
+impl GracefulWatcher {
     /// Watch a future for graceful shutdown,
     /// returning a wrapper that can be awaited on.
-    pub fn watch<C, F>(
-        &self,
-        conn: C,
-        cancel: F,
-    ) -> Result<GracefulFuture<C::Error>, GracefulShutdownError<C>>
+    pub fn watch<'a, C, F>(&'a self, conn: C, cancel: F) -> GracefulFuture<C::Error>
     where
-        C: GracefulConnection + Send + Sync + 'static,
-        F: Future + Send + Sync + 'static,
+        C: GracefulConnection + Send + Sync + 'a,
+        F: Future + Send + Sync + 'a,
     {
-        if self
-            .state
-            .shutdown
-            .load(std::sync::atomic::Ordering::SeqCst)
-        {
-            return Err(GracefulShutdownError { conn });
-        }
-        Ok(GracefulFuture {
+        self.state.subscribe();
+        GracefulFuture {
             future: Box::pin(async move {
                 let mut conn = pin!(conn);
                 let mut cancel = pin!(cancel.fuse());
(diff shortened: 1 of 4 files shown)


seanmonstar:
First of all, the work in here looks really nice!

> I do not take a generic future but instead implement a trait myself that I implement for the http1 conn, http2 conn and auto conn (when feature is enabled), so that I can do the dance for the user

I think this is probably the best one. You can look at how [hyper 0.14.x did it](https://github.com/hyperium/hyper/blob/0.14.x/src/server) for the internal future logic.

Some other thoughts:

- If you remove the `Clone` impl, then I think you could also remove the `Result` from the `watch()` method, since you know it could never have started shutdown on a clone.

GlenDC:
Great to hear @seanmonstar will look to finish this PR for a first non-draft review this week then, based on this feedback.

GlenDC:
@seanmonstar the clone can only be removed if you only ever want one shutdown? Unless you start stacking shutdowns?
I was respecting for now the desired API at https://github.com/hyperium/hyper/issues/2862, is that no longer desired then?

GlenDC:
@seanmonstar and @davidpdrsn I know have a working example: <https://github.com/hyperium/hyper-util/blob/8f24314cf040faba15190ff5b75652d3e4c70e52/examples/server_graceful.rs>

You can play with it yourself.

Figured that doing the example already would help me iron out some details myself.

- I now no longer require Clone and thus the watch function also requires no Result (moved it out into a watcher, which keeps a ref count of itself);
- I no longer require the connection / future to be static, but instead to live as long as the watcher. This makes it all work nicely.

Example works and AFAIK the design is now final and ready for your review.

Once you both approve of the design and we worked out some comments you might have, i can finish this off by adding also some unit tests.

GlenDC:
@seanmonstar and @davidpdrsn , this PR is now ready for "final" review, I am finished.

Updates since last time:

- added tests
- improved bits and bobs
- dropped dependency on tokio, by not using tokio::select
- no longer require the cancel func in watch, we now fully shutdown based on the single shutdown call

There's also an example that you can play with:

```
cargo run --features="tokio server-graceful server-auto" --example server_graceful
```

Looking forward to the review. As it's my first PR to this organisation I do expect there might be some more back and forward needed. You tell me, as I do think it's ready but 🤷 who knows.

GlenDC:
@seanmonstar is there work to be done here or something I can do to help make this progress. Also if there's something else you have in mind I can look at next, I could take a look at it. The hyper issue seemed interesting but I think I would like to also start contributing a bit to the h2/h3 crates as a way to motivate myself to get more familiar with these.

Anything you have in mind?

seanmonstar:
Sorry I haven't had the time to review again. I was just out for a week escaping the winter.

@programatik29 @dswij @tottoto any of you want to take a look? (Relevant issue and discussion: https://github.com/hyperium/hyper/issues/2862)

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+    sync::{atomic::AtomicUsize, Arc, Mutex},
+    task::{self, Poll, Waker},
+};
+
+/// A graceful shutdown watcher
+pub struct GracefulShutdown {
+    // state used to keep track of all futures that are being watched
+    state: Arc<GracefulState>,
+    // state that the watched futures to know when shutdown signal is received
+    future_state: Arc<GracefulState>,
```
dswij, line 24: IIUC, the future_state here is to keep track of all the futures. What does the `state` do for GracefulShutdown here?

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+    }
+
+    /// Get a graceful shutdown watcher
+    pub fn watcher(&self) -> GracefulWatcher {
+        self.state.subscribe();
+        GracefulWatcher {
+            state: self.state.clone(),
+            future_state: self.future_state.clone(),
+        }
+    }
```
dswij, line 61: We can spawn multiple `GracefulWatcher` from a single `GracefulShutdown`, so is it accurate to say that the states in GracefulShutdown is for every watcher and futures?

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+            None => waker_list.insert(waker),
+        });
+
+        Poll::Pending
+    }
+}
+
+struct GracefulWaiterState {
+    state: Arc<GracefulState>,
+    key: Option<usize>,
```
dswij, line 229: What does the `key` here do?

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+
+        let mut waker_list = state.state.waker_list.lock().unwrap();
+
+        if state
+            .state
+            .counter
+            .load(std::sync::atomic::Ordering::SeqCst)
+            == 0
+        {
+            // check again in case of race condition
```
dswij, line 210: What's the scenario for race condition here?

src/server/graceful.rs
GlenDC: IIUC = If I understand Correctly, didn't know that was a thing.

src/server/graceful.rs
GlenDC: We first check the counter. That one is zero only if shutdown is already triggered. We then wait for a lock. However it is possible that that we had to wait for the lock while something else was holding the lock to awake the ones waiting for shutdown.

If we do not check again once we have the lock it is possible that the counter in meanwhile is zero. If so we either way need to trigger shutdown here or else we might not be woken up.

This was discovered while developing the origin code on which this PR was inspired (https://github.com/plabayo/tokio-graceful, a crate of mine), where loom helped me to find that issue.

src/server/graceful.rs
GlenDC: Is that a question or you want me to comment it? It is the insertion key in the waker list, important for it to be removed again.

src/server/graceful.rs
GlenDC: Yes. That is accurate, they all lead to the same watcher. Once you shutdown however you consume the `GracefulShutdown` and no new graceful watchers can be created.

src/server/graceful.rs
GlenDC: Naming is hard though. I am essentially using the same concept in two directions.

- the "state" is to keep track of all watchers active. When the `GracefulShutdown` turns into a `GracefulWaiter` upon `self::shutdown()` it uses it to check when the counter is 0 and thus when shutdown is finished. While it is still a `GracefulShutdown` it clones this state and patches it to the `GracefulWatchers` (the ones watching the "graceful" connections) to help to do its bookkeeping using this counter
- the "future_state" is only ever once incremented (it is the counter of the `GracefulShutdown` itself, of which there is only ever one. And it is used by the "graceful" connections to know if a cancel is triggered. This is essentially "reusing" the `GracefulWaiter` logic that was already created for the other direction.

So no probably due to the fault of my part (not sure if bad documentation, naming or code) it seemed that you didn not understand it correctly. As it is the `state` which is there to keep track of all the futures (futures as in graceful connections).

GlenDC:
Recreated the PR as https://github.com/hyperium/hyper-util/pull/108. Sorry for the confusion @seanmonstar and @dswij .

Github does say the vague "bad things will happen" msg when deleting repo, but might be more useful if it can actually say that only and with specifics. As I didn't realise this PR was still attached to that repo. Thought it was a fork from an older idea.

PR #91, opened 2024-01-16
fix: unpinned patch version in dependencies

ikrivosheev:
Hello! I unpinned patch version in dependencies.


davidpdrsn:
This is not necessary with Cargo. See https://github.com/tokio-rs/axum/pull/2346 for more details.

ikrivosheev:
@davidpdrsn thank you for explain)

PR #92, opened 2024-01-16
Fix graceful shutdown for Http(1|2)Builder

k0nserv:
Because `Http1Builder::serve_connection` and
`Http2Builder::serve_connection` didn't return `Connection` it was
impossible to use them in combination with `Connection::graceful_shutdown`.

By returning the Connection future rather than awaiting it, this becomes
possible.


seanmonstar:
This was done in another PR.

PR #93, opened 2024-01-24
feat: impl From<Http1|Http2> for auto::Builder

seanmonstar:
The motivation is I wanted a way to get back to the main `Builder`, and this seemed to do it. Does it make sense?

cc @davidpdrsn @programatik29

### DIFF
--- a/src/server/conn/auto.rs
+++ b/src/server/conn/auto.rs
@@ -113,6 +113,19 @@ impl<E> Builder<E> {
         }
     }
 }
+
+impl<'a, E> From<Http1Builder<'a, E>> for &'a mut Builder<E> {
+    fn from(h1: Http1Builder<'a, E>) -> &'a mut Builder<E> {
+        h1.inner
+    }
+}
+
+impl<'a, E> From<Http2Builder<'a, E>> for &'a mut Builder<E> {
+    fn from(h2: Http2Builder<'a, E>) -> &'a mut Builder<E> {
+        h2.inner
+    }
+}
+
 #[derive(Copy, Clone)]
 enum Version {
     H1,


dswij (approved):

tottoto (approved):

martinetd:
Note on this: `builder.http()` returns `Http2Builder` so using `into()`  immediately after it works, but setting any option will return a `&mut Http2Builder` which we cannot get back from, so I'm not sure how useful that is in practice unless we also make the methods not take refs.

Either way the Http[12]Builder only take a mut ref to Builder, so it's possible to change http2 options and go back which I hadn't realized in #113 ; so something like this works:
```
                    let mut builder = auto::Builder::new(TokioExecutor::new());
    
                    builder
                        .http2()
                        .max_header_list_size(4096);
        
                    builder
                        .serve_connection_with_upgrades(stream, service_fn(hello))
                        .await;
```
It's a bit less pretty than just chaining operators, but that means there probably isn't any hard need for this.

cratelyn:
@seanmonstar i've been going through the open pull requests and wanted to ask about this one. do you think that we should go forward with this? #98 proposes another alternative way of solving this problem, but i'm not sure if i have enough context handy to have a clear preference between the two.

we could also consider closing both and filing an issue. what do you think?

seanmonstar:
eh, I don't know. i think i was able to get around the problem I ran into back then. so i can kill this.

(i'm not sure the current auto builder is as nice as possible to use, but what can ya do)

cratelyn:
> eh, I don't know. i think i was able to get around the problem I ran into back then. so i can kill this.
> 
> (i'm not sure the current auto builder is as nice as possible to use, but what can ya do)

yeah, that makes sense to me too. in that case, i think #98 may be worth closing as well, would you agree?

PR #95, opened 2024-01-29
feat: Add a `Error::is_connect()` method

jakubadamw:
It will be useful for reqwest's upgrade to hyper v1.

### DIFF
--- a/src/client/legacy/client.rs
+++ b/src/client/legacy/client.rs
@@ -1516,6 +1516,10 @@ impl StdError for Error {
 }
 
 impl Error {
+    pub fn is_connect(&self) -> bool {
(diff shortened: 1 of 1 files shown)


src/client/legacy/client.rs
```diff
@@ -1516,6 +1516,10 @@ impl StdError for Error {
 }
 
 impl Error {
+    pub fn is_connect(&self) -> bool {
```
seanmonstar, line 1519:
```suggestion
    /// Returns true if this was an error from `Connect`.
    pub fn is_connect(&self) -> bool {
```

seanmonstar (approved):

PR #103, opened 2024-03-01
fix: enable socket2 only for client-legacy

elpiel:
`socket2` dependency is only used in the client-legacy.
Current approach had the disadvantage of enabling socket2 even if you don't need it, e.g. a platform that socket2 doesn't support like `wasm-unknown-unknown`


seanmonstar (approved):
Thanks!

PR #104, opened 2024-03-01
chore: make futures-util optional

tottoto:
Makes `futures-util` dependency optional.


tottoto:
This seems not work without a breaking change. Taken over by #105.

PR #105, opened 2024-03-02
feat(server-auto): make http1 and http2 feature optional

tottoto:
Depends on https://github.com/hyperium/hyper/pull/3591. 

This is a breaking change. Uses `server-auto` feature to enable server-auto implementation to treat dependencies finely same as client and client-legacy do.


cratelyn:
superseded by #264.

PR #106, opened 2024-03-07
tests: add teest for graceful shutdown

mladedav:
I've added a test that checks that a connection that has not written anything is closed in a timely manner.

I've checked that the test would have failed with a timeout before this has been fixed.

Closes [#3582](https://github.com/hyperium/hyper/issues/3582)


dswij (approved):
Nice, thank you so much! @mladedav

PR #108, opened 2024-03-13
#2862: GracefulShutdown utility 

GlenDC:
Closes https://github.com/hyperium/hyper/issues/2862

Adds a GracefulShutdown utility which:

- is runtime agnostic
- can work with any future
- respects the API as requested in the feature request issue (#2862), taken also into account comments and ideas from this PR;

recreation of https://github.com/hyperium/hyper-util/pull/86

Sorry, created my original repo under the plabayo org, didn't realise that that PR was still attached to it.

### DIFF
--- a/examples/server_graceful.rs
+++ b/examples/server_graceful.rs
@@ -1,0 +1,74 @@
+use bytes::Bytes;
+use std::convert::Infallible;
+use std::pin::pin;
+use std::time::Duration;
+use tokio::net::TcpListener;
+use tokio::sync::broadcast::channel;
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let listener = TcpListener::bind("127.0.0.1:8080").await?;
+
+    let graceful = hyper_util::server::graceful::GracefulShutdown::new();
+    let mut ctrl_c = pin!(tokio::signal::ctrl_c());
+    let (shutdown_tx, _) = channel(1);
+
+    loop {
+        tokio::select! {
+            conn = listener.accept() => {
+                let (stream, peer_addr) = match conn {
+                    Ok(conn) => conn,
+                    Err(e) => {
+                        eprintln!("accept error: {}", e);
+                        tokio::time::sleep(Duration::from_secs(1)).await;
+                        continue;
+                    }
+                };
+                eprintln!("incomming connection accepted: {}", peer_addr);
+                let mut shutdown_rx = shutdown_tx.subscribe();
+
+                let stream = hyper_util::rt::TokioIo::new(Box::pin(stream));
+                let cancel = async move {
+                    let _ = shutdown_rx.recv().await;
+                    eprintln!("connection shutdown triggered {}", peer_addr);
+                };
+                let watcher = graceful.watcher();
+
+                tokio::spawn(async move {
+                    let server = hyper_util::server::conn::auto::Builder::new(hyper_util::rt::TokioExecutor::new());
+                    let conn = server.serve_connection_with_upgrades(stream, hyper::service::service_fn(|_| async move {
+                            tokio::time::sleep(Duration::from_secs(5)).await;  // emulate slow request
+                            let body = http_body_util::Full::<Bytes>::from("Hello World!".to_owned());
+                            Ok::<_, Infallible>(http::Response::new(body))
+                        }));
+
+                    let conn = watcher.watch(conn, cancel);
+
+                    if let Err(err) = conn.await {
+                        eprintln!("connection error: {}", err);
+                    }
+                    eprintln!("connection dropped: {}", peer_addr);
+                });
+            },
+
+            _ = ctrl_c.as_mut() => {
+                drop(listener);
+                eprintln!("Ctrl-C received, starting shutdown");
+                    break;
+            }
+        }
+    }
+
+    let _ = shutdown_tx.send(());
+
+    tokio::select! {
+        _ = graceful.shutdown() => {
+            eprintln!("Gracefully shutdown!");
+        },
+        _ = tokio::time::sleep(Duration::from_secs(10)) => {
+            eprintln!("Waited 10 seconds for graceful shutdown, aborting...");
+        }
+    }
+
+    Ok(())
+}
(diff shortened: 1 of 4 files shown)


GlenDC:
Maybe best to start a new discussion here @dswij if you have further questions on the ones raised in the original PR.

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+
+        if state
+            .state
+            .counter
+            .load(std::sync::atomic::Ordering::SeqCst)
+            == 0
+        {
+            // check again in case of race condition
+            return Poll::Ready(());
+        }
```
tottoto, line 212: I think that user want to know whether this was polled successfully or not (in this case, race condition might occur). The `Output` type can explain the result instead of just empty tuple?

If so, would it be possible to test for this race condition?

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+            None => waker_list.insert(waker),
+        });
+
+        Poll::Pending
+    }
+}
+
+struct GracefulWaiterState {
+    state: Arc<GracefulState>,
+    key: Option<usize>,
```
tottoto, line 229: Does this `key` describe the cursor of the state?

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+where
+    C: GracefulConnection,
+    F: Future,
+{
+    type Output = C::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
+        let mut this = self.project();
+        if !*this.cancelled {
+            if let Poll::Ready(_) = this.cancel.poll(cx) {
```
tottoto, line 279: `is_ready` can be used, and then this nested `if` statement can be flattened.

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+    E: hyper::rt::bounds::Http2ServerConnExec<S::Future, B>,
+{
+    type Error = Box<dyn std::error::Error + Send + Sync>;
+
+    fn graceful_shutdown(self: Pin<&mut Self>) {
+        crate::server::conn::auto::UpgradeableConnection::graceful_shutdown(self);
+    }
+}
+
+mod private {
```
tottoto, line 368: According to the existing hyper projects, the conventional module name seems `sealed`.

examples/server_graceful.rs
```diff
@@ -0,0 +1,65 @@
...
+                        eprintln!("connection error: {}", err);
+                    }
+                    eprintln!("connection dropped: {}", peer_addr);
+                });
+            },
+
+            _ = ctrl_c.as_mut() => {
+                drop(listener);
+                eprintln!("Ctrl-C received, starting shutdown");
+                    break;
```
tottoto, line 50: Indentation?

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+    sync::{atomic::AtomicUsize, Arc, Mutex},
+    task::{self, Poll, Waker},
+};
+
+/// A graceful shutdown watcher
+pub struct GracefulShutdown {
+    // state used to keep track of all futures that are being watched
+    state: Arc<GracefulState>,
+    // state that the watched futures to know when shutdown signal is received
+    future_state: Arc<GracefulState>,
```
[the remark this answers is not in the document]
GlenDC: Any suggestions?

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+}
+
+impl Debug for GracefulShutdown {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("GracefulShutdown").finish()
+    }
+}
```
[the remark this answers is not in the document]
GlenDC: Sure can do!

src/server/graceful.rs
```diff
@@ -0,0 +1,620 @@
...
+impl GracefulShutdown {
+    /// Create a new graceful shutdown watcher
+    pub fn new() -> Self {
+        Self {
+            state: Arc::new(GracefulState {
+                counter: AtomicUsize::new(0),
+                waker_list: Mutex::new(Slab::new()),
+            }),
+            future_state: Arc::new(GracefulState {
+                counter: AtomicUsize::new(1),
```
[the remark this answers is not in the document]
GlenDC: It's because this is the property used by the tasks to know if a graceful cancellation is requested. As such it starts at 1 as the main struct is still not cancelled. It goes to 0 once it is being cancelled. It looks a bit odd but it's basically reusing the same `GracefulState` struct that was already used for the other direction

src/server/graceful.rs
GlenDC: sure I have no strong opinion on this, you want me to change it?

src/server/graceful.rs
GlenDC: ok will check it out

src/server/graceful.rs
GlenDC: The key is set in case it already has a waker registered for it. If so then we replace it instead of generating a new one in case it got polled again before woken up

src/server/graceful.rs
GlenDC: Not sure I can follow? The race condition is tested for in my original https://github.com/plabayo/tokio-graceful library, but none of that testing infra is in place here. Either way for the user it shouldn't matter as this is an implementation detail? All that matters is that this thing Polls to ready in case the waiter is done waiting. The race conditon can happen, but because of this check it is caught and so all is good after all. Not sure that I can follow why you would need as a user to see the distinction?

seanmonstar:
I had this thought: while having an optimized data structure is a nice-to-have, what if this internally used what hyper v0.14.x used, just a pair of mpsc and watch channels? I think they're pretty easy to reason about, and we don't have to worry about subtle bugs. We could probably merge _that_ sooner, and then a specialized data structure could be a follow up.

I could also do that, using this PR as a starting off point...

GlenDC:
All the same to me @seanmonstar . Given that you are busy enough as is, I will finish it myself.

Before that, can you please confirm that the above approach proposal of yours (mpcs + watch channels) is what you want. I’ll update this PR code once I have that confirmation.

seanmonstar:
I do think porting those internals is a good starting place. The public API you have here is great for users.

GlenDC:
Superseded by https://github.com/hyperium/hyper-util/pull/127.

PR #111, opened 2024-03-17
feat: add `{http1,http2}_only` for auto conn

dswij:
Closes https://github.com/hyperium/hyper/issues/3511

This PR adds an option to choose http1 or http2 only for auto connections.


seanmonstar:
@sfackler would this solve your use case?

src/server/conn/auto.rs
```diff
@@ -101,6 +105,26 @@ impl<E> Builder<E> {
         Http2Builder { inner: self }
     }
 
+    /// Only accepts HTTP/2
+    ///
+    /// Does not do anything if used with [`serve_connection_with_upgrades`]
+    #[cfg(feature = "http2")]
+    pub fn http2_only(mut self) -> Self {
```
cxw620, line 112: What about adding a param controlling if enable http2_only or not? So that we can just write such codes(example codes from Line 1068-1075):
```rust
                    auto::Builder::new(TokioExecutor::new())
                        .http1_only(h1_only)
                        .http2_only(h2_only)
                        .serve_connection(stream, service_fn(hello))
                        .await;
```
instead of
```rust
                    let mut builder = auto::Builder::new(TokioExecutor::new());
                    if h1_only {
                        builder = builder.http1_only();
                    } else if h2_only {
                        builder = builder.http2_only();
                    }

                    builder.serve_connection(stream, service_fn(hello)).await;
```

As far as I'm concerned, the latter snippet lacks elegance.

dswij: You're right in terms of ergonomics, but the first example somewhat suggests that we expect the user to turn the knob for both of them.

With the second example, while not elegant, does suggest that you set only *either* at any given time.

cxw620: > You're right in terms of ergonomics, but the first example somewhat suggests that we expect the user to turn the knob for both of them.
> 
> With the second example, while not elegant, does suggest that you set only _either_ at any given time.

I must admit that I agree with you about avoiding the user's mistake of setting `http1_only` or `http2_only` multiple times. After all, there are always users who don't read the documentation carefully, or mistakenly do so. 

However, considering that `hyper-util` is a relatively low-level library, should we provide more flexibility to users and let themselves be responsible for not setting `http1_only` or `http2_only` multiple times? BTW, in my opinion, manual runtime panic in release mode doesn't seem quite appropriate, since it's actually not an unrecoverable error, consider using `debug_assert`?

tisonkun, line 112: I'd instead suggest:

```rust
    pub fn http2_version(mut self, v: Option<Version>) -> Self
```

... and expose `Version` struct. Otherwise, it's panicable when you call with `http1_only = true and http2_only = true`, or `self.http1_only().http2_only()`. It's hard to tell if a passed builder is already configured. `version` setter override the configured value, at least better than assert! not set.

tisonkun (approved):
Share my opinion on the style issue. In general, I'm OK with either way and let's try to move forward.

src/server/conn/auto.rs
mcmah309: As @tisonkun pointed out, seems like a pretty straight forward case for exposing the enum. Why introduce the possibility of an error or require conditional logic.
```rust
async fn start_server(version: Option<Version>) -> SocketAddr {
    ...
    let mut builder = auto::Builder::new(TokioExecutor::new());
    builder = match version {
        Some(version) => builder.http_version(version),
        None => {}
    };
    builder.serve_connection(stream, service_fn(hello)).await;
```
I don't think `http_version` should accept an `Option` though, since `None` is already the default. But if you do allow, then
```rust
async fn start_server(version: Option<Version>) -> SocketAddr {
    ...
     auto::Builder::new(TokioExecutor::new())
            .http_version(version)
            .serve_connection(stream, service_fn(hello))
            .await;
```

rakshith-ravi:
Hey. Just wanted to check in to see if there's anything that I can do to help push this forward?

src/server/conn/auto.rs
seanmonstar: Accepting an `Option` would make it even easier for those who are doing something like reading a config file or similar. If not present, the value is `None`. It could be `http_version(impl Into<Option<..>>)`.

But.

But if we add that, let's make it a separate PR. I think there's details to work out, and it both ways of configuring can exist. We don't need to block this PR on that. What's in here is good.

seanmonstar (approved):
Thanks dswij! Can be merged after conflicts are resolved.

tisonkun:
@seanmonstar I send a follow-up to resolve the conflict at https://github.com/hyperium/hyper-util/pull/124.

It retains @dswij's credit. Hopefully we can move forward and my major target is to help https://github.com/hyperium/tonic/pull/1670 done.

tisonkun:
@seanmonstar This patch has some uncompleted part. I fix it up at https://github.com/hyperium/hyper-util/pull/124/commits/582cd9fcb926651f601e9d2423c9e54f568d0cf5.

PR #114, opened 2024-04-09
fix: disable default features for dep "tower"

hasezoey:
`hyper-util` implicitly depends on `tower`'s feature `log` through `default`s, which should not be necessary. They have [long since removed the defaults](https://github.com/tower-rs/tower/commit/5064987ffec4420ef6ea6de649606286370e42bc), but never actually released a new version with that fix.

I have only found this because i upgraded to reqwest 0.12 (from 0.11), which suddenly spammed the logs (which were not setup for tracing yet) with tracing output, but those logs could somehow not be disabled, and finding the cause was quite annoying. This all happened because `tower`'s `log` feature sets `tracing/log`.


seanmonstar (approved):

CryZe:
Can we have a release with this? The log spam is forcing me to stay on 0.11 for the time being.

PR #117, opened 2024-04-12
feat: impl Connection for TokioIo

kvinwang:
Without this implementation, when a downstream crate wants to implement a custom `HttpConnector`, it would have to wrap up an extra layer over `TokioIo<TcpStream>`. Consequently, this would result in a forwarding implementation of `hyper::rt::{Read, Write}` for the new type.

See [here](https://github.com/kvinwang/wapo/commit/9e748d401128101efe05098f740a65b1549771ca) for more detail.

### DIFF
--- a/src/client/legacy/connect/http.rs
+++ b/src/client/legacy/connect/http.rs
@@ -438,12 +438,10 @@ where
     }
 }
 
-impl Connection for TokioIo<TcpStream> {
+impl Connection for TcpStream {
     fn connected(&self) -> Connected {
         let connected = Connected::new();
-        if let (Ok(remote_addr), Ok(local_addr)) =
-            (self.inner().peer_addr(), self.inner().local_addr())
-        {
+        if let (Ok(remote_addr), Ok(local_addr)) = (self.peer_addr(), self.local_addr()) {
             connected.extra(HttpInfo {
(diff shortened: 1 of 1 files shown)


src/rt/tokio.rs
```diff
@@ -212,6 +212,16 @@ where
...
+#[cfg(feature = "client-legacy")]
+impl<T> crate::client::legacy::connect::Connection for TokioIo<T>
+where
+    T: crate::client::legacy::connect::Connection,
+{
+    fn connected(&self) -> crate::client::legacy::connect::Connected {
+        self.inner.connected()
+    }
+}
+
```
tottoto, line 224: I would think especially implementations regarding legacy would be better to be located in legacy module.

tottoto:
Should some document about this be needed.

tottoto:
Thanks! Sounds reasonable to me. I left some comments.

src/rt/tokio.rs
kvinwang: OK. Moved it into the lagacy module.

tottoto (approved):

dswij (approved):

tottoto:
Thank you!

PR #123, opened 2024-05-10
chore(client): move legacy client builder http2 timer config to http2 feature

tottoto:
This config API do nothing without `http2` feature. This might be confusing.

As the other `http2` specific config APIs are behind `http2` feature and have `http2_` prefix, unifying it to them will make it easier to be understood.

This is a breaking change.


cratelyn:
thank you for opening this. i agree that this is probably more correct, but i think at this stage we are trying to avoid breaking changes to the legacy client. apologies that this took so long to be reviewed.

PR #124, opened 2024-05-18
feat: add {http1,http2}_only for auto conn

tisonkun:


src/server/conn/auto.rs
```diff
@@ -1011,10 +1106,16 @@ mod tests {
...
-                        .http2()
+                    let mut builder = auto::Builder::new(TokioExecutor::new());
+                    if h1_only {
+                        builder = builder.http1_only();
+                    } else if h2_only {
+                        builder = builder.http2_only();
+                    }
+
+                    builder
                         .max_header_list_size(4096)
```
martinetd, line 1117: max_header_list_size only works on Http2Builder, so you need to keep the `.http2()` here, which as far as I understand will not prevent http1 tests from working (the Http2Builder is only used to set http2 options, and doesn't imply http2_only)

Logically it's easier to read something that'd do the other way around (`builder.http2().max_header_list_size(...)` first then get back an auto Builder and run http[12]_only and then serve it), so perhaps something like this instead?
```
                    let mut builder: auto::Builder<_> = auto::Builder::new(TokioExecutor::new());
    
                    builder
                        .http2()
                        .max_header_list_size(4096);
        
                    if h1_only {
                        builder = builder.http1_only();
                    } else if h2_only {
                        builder = builder.http2_only();
                    }
        
                    builder
                        .serve_connection_with_upgrades(stream, service_fn(hello))
                        .await;
```

tisonkun: Thank you! Addressed.

tisonkun:
@seanmonstar make sense. Thanks for managing this :D

tisonkun:
For failing CI, I believe it's because my local verification miss something in CI that is lagged due to every CI run requires approval.

I'll try to make other contributions to gain the credit run CI without explicit approval :P

PR #126, opened 2024-05-25
Config rust-version at Cargo.toml

tisonkun:
Cargo.toml supports MSRV - https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

cc @seanmonstar

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,8 +38,14 @@ jobs:
     needs: [style]
     strategy:
       matrix:
-        rust: [ stable, beta, nightly ]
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -28,75 +28,51 @@ jobs:
...
       matrix:
-        rust:
-          - stable
-          - beta
-          - nightly
-
-        os:
-          - ubuntu-latest
-          - windows-latest
-          - macOS-latest
```
seanmonstar, line 51: Can we keep all these bullet points? I find it easier to scan the matrix with my eyes that way.

seanmonstar (approved):

PR #127, opened 2024-05-27
feat: add GracefulShutdown helper

seanmonstar:
This adds a new `hyper_util::server::graceful::GracefulShutdown` helper, which can be used to watch connections, and then trigger a graceful shutdown and await all connections finishing.

Continues off of #108. cc @glendc

Closes https://github.com/hyperium/hyper/issues/2862

### DIFF
--- a/src/server/graceful.rs
+++ b/src/server/graceful.rs
@@ -1,0 +1,450 @@
+//! Utility to gracefully shutdown a server.
+//!
+//! This module provides a [`GracefulShutdown`] type,
+//! which can be used to gracefully shutdown a server.
+//!
+//! See <https://github.com/hyperium/hyper-util/blob/master/examples/server_graceful.rs>
+//! for an example of how to use this.
+
+use std::{
+    fmt::{self, Debug},
+    future::Future,
+    pin::Pin,
+    task::{self, Poll},
+};
+
+use pin_project_lite::pin_project;
+use tokio::sync::watch;
+
+/// A graceful shutdown utility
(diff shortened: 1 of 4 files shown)


src/server/graceful.rs
```diff
@@ -0,0 +1,447 @@
...
+
+impl GracefulShutdown {
+    /// Create a new graceful shutdown helper.
+    pub fn new() -> Self {
+        let (tx, rx) = watch::channel(());
+        Self { tx, rx }
+    }
+
+    /// Wrap a future for graceful shutdown watching.
+    pub fn watch<C: GracefulConnection>(&self, conn: C) -> impl Future<Output=C::Output> {
```
seanmonstar, line 34: I'm actually thinking of making this `&mut self`, even though it's not _required_, but it would allow us to change internals.

GlenDC:
Thx. I currently have no time. So I’ll review instead :3

GlenDC (approved):
I have a feeling you're not ready yet. But this does look a whole lot simpler indeed!
Not surprising of course, as there is a pay off. Fully respect the approach though, and pays of it self indeed. Thx for picking up the work, and all your hard work in general!!!

src/server/graceful.rs
GlenDC: What would be a hypoteical practical case where you need &mut in a scenario like this?

seanmonstar: If we changed from using a channel to something that mutates `self` non-atomically. It is hard to think of how that could work...

PR #131, opened 2024-06-18
fix: remove unnecesary receiver field from GracefulShutdown

ramiroaisen:
This PR removes the unnecesary `watch::Receiver<()>` field from the `GracefulShutdown` struct.
The receiver field is unnecesary since a Receiver can be obtained with `Sender::subscribe()`.

This reduces the size of the struct from 24 to 8.

After this is merged we can think if its worth it to enable `Clone` of the `GracefulShutdown`, which will produce a more flexible api for consumers.


seanmonstar (approved):
Thanks!

PR #134, opened 2024-07-03
fix: run dns resolution in the same tracing-span as the caller

joshka:
This makes it possible to trace the entire request, including DNS
resolution.

The use case for this is to be able to suppress specific requests from
being traced in a situation where the request is used to transmit logs
to a remote server. This would result in an infinite loop of logs being
sent to the remote server. https://github.com/tokio-rs/tokio/issues/6659
has more details.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -4,22 +4,9 @@ use http_body_util::Empty;
 use hyper::Request;
 use hyper_util::client::legacy::{connect::HttpConnector, Client};
 use tracing::{info_span, Instrument};
-use tracing_subscriber::{
-    fmt::{self, format::FmtSpan},
-    prelude::*,
-    EnvFilter,
-};
 
 #[tokio::main(flavor = "current_thread")]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let fmt_layer = fmt::layer()
-        .with_span_events(FmtSpan::CLOSE) // show time elapsed in spans
-        .with_timer(fmt::time::Uptime::default());
-    tracing_subscriber::registry()
-        .with(EnvFilter::from_default_env())
-        .with(fmt_layer)
-        .init();
-
     let url = match env::args().nth(1) {
         Some(url) => url,
         None => {
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -3,7 +3,6 @@ use std::env;
 use http_body_util::Empty;
 use hyper::Request;
 use hyper_util::client::legacy::{connect::HttpConnector, Client};
-use tracing::{info_span, Instrument};
 
 #[tokio::main(flavor = "current_thread")]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -29,8 +28,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .uri(url)
         .body(Empty::<bytes::Bytes>::new())?;
 
-    let span = info_span!("request", uri = %req.uri());
-    let resp = client.request(req).instrument(span).await?;
+    let resp = client.request(req).await?;
 
     eprintln!("{:?} {:?}", resp.version(), resp.status());
     eprintln!("{:#?}", resp.headers());
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -3,9 +3,23 @@ use std::env;
 use http_body_util::Empty;
 use hyper::Request;
 use hyper_util::client::legacy::{connect::HttpConnector, Client};
+use tracing::{info_span, Instrument};
+use tracing_subscriber::{
+    fmt::{self, format::FmtSpan},
+    prelude::*,
+    EnvFilter,
+};
 
 #[tokio::main(flavor = "current_thread")]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let fmt_layer = fmt::layer()
+        .with_span_events(FmtSpan::CLOSE) // show time elapsed in spans
+        .with_timer(fmt::time::Uptime::default());
+    tracing_subscriber::registry()
+        .with(EnvFilter::from_default_env())
+        .with(fmt_layer)
+        .init();
+
     let url = match env::args().nth(1) {
         Some(url) => url,
         None => {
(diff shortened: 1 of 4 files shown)


joshka:
Pingw

src/client/legacy/connect/dns.rs
```diff
@@ -118,7 +118,9 @@ impl Service<Name> for GaiResolver {
     }
 
     fn call(&mut self, name: Name) -> Self::Future {
+        let current_span = tracing::Span::current();
         let blocking = tokio::task::spawn_blocking(move || {
+            let _enter = current_span.enter();
```
seanmonstar, line 123: The only question I have is since the call below is blocking, does entering the span the whole time mess with measurements?

cc @hawkw

joshka: That's a good thing though right? Is there a reason that the the timing of a http request shouldn't include DNS lookup time. Are you concerned with existing measurement metrics looking bad because they now include this time?

I think it might be worth creating a child debug span here for the DNS resolution (done in 0f690ae). This allows metrics for just dns resolution to be visible when the debug span level is enabled:

debug enabled (dns time is not included in the main span):
```shell
RUST_LOG=debug cargo run --example=client --features=full http://example.com
```
image


debug disabled (dns time is included in the main span):
```shell
RUST_LOG=info cargo run --example=client --features=full http://example.com
```
image

seanmonstar: Ah ok, cool. I didn't explain what I meant well, I was worried that since the span would be in a "blocking" task, I worried that the overall span could appear to be blocking a bit, if you looked in tokio-console, for instance.

If that's not the case, I'm happy to merge this, with the extra tracing-subscriber stuff pulled back out of the example.

joshka: There is a single TRACE event that marks the blocking task:
```
   0.011094292s DEBUG request{uri=http://example.com/}:resolve{host=example.com}: hyper_util::client::legacy::connect::dns: resolving host="example.com"
   0.015239834s DEBUG request{uri=http://example.com/}:resolve{host=example.com}: hyper_util::client::legacy::connect::dns: close time.busy=4.15ms time.idle=227µs
   0.015319375s TRACE request{uri=http://example.com/}:runtime.spawn{kind=blocking task.name= task.id=5 fn=<hyper_util::client::legacy::connect::dns::GaiResolver as tower_service::Service<hyper_util::client::legacy::connect::dns::Name>>::call::{{closure}} loc.file="src/client/legacy/connect/dns.rs" loc.line=122 loc.col=24}: tokio::task::blocking: close time.busy=4.25ms time.idle=155µs
   0.015470667s TRACE request{uri=http://example.com/}:runtime.resource{concrete_type="Sleep" kind="timer" loc.file="src/client/legacy/connect/http.rs" loc.line=603 loc.col=28}: runtime::resource::state_update: duration=301 duration.unit="ms" duration.op="override"
   0.015578875s DEBUG request{uri=http://example.com/}: hyper_util::client::legacy::connect::http: connecting to 93.184.215.14:80
...
```
which results in a blocking task showing up briefly in tokio console:
image

But the blocked time is allocated to the blocked task, which is probably correct in that it represents the reality of what's actually happening there.

joshka: Removed the tracing_subscriber btw.

examples/client.rs
```diff
@@ -28,7 +29,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .uri(url)
         .body(Empty::<bytes::Bytes>::new())?;
 
-    let resp = client.request(req).await?;
+    let span = info_span!("request", uri = %req.uri());
+    let resp = client.request(req).instrument(span).await?;
```
seanmonstar, line 33: Could we remove this too?

I agree it's useful when debugging, but I worry it distracts from the initial usage of someone just wanting to get the client working.

joshka: yep 2bce070..5f7432e

seanmonstar (approved):
Thank you!

arpadav:
@seanmonstar 

This is causing: https://github.com/tokio-rs/tracing/issues/3223#issuecomment-2771038993 

Took me a while to track this down. But reverting this change solves this issue.

If you are interested in resolving this, I can think of 2 courses of action:

1. Reverting this change
2. Keeping this change, but add `#[cfg(feature = "tracing")]` to all `tracing` calls, and then remove  `tracing` in https://github.com/hyperium/hyper-util/blob/e7b396628006f5c9279b4f93a6a3ca8a94a2548d/Cargo.toml#L62

(2) This would break people depending on `tracing` when using the `client` feature, however. A manual update of their Cargo.toml from `hyper-util = { version = "~0.1", features = ["client"] }` -> `hyper-util = { version = "~0.1", features = ["client", "tracing"] }` would be required. Is this desired? Its a quick fix I am more than happy to put in a PR for.

joshka:
> At this point, I'm happy to accept a PR reverting the whole thing. Providing a slightly more accurate trace isn't worth panics.

I think this would be the right short term approach, but wrong as a long term idea. I think it's important if this is causing panics to revert first and make it possible to work on a replacement that doesn't.

I haven't dug into the tracing bug to understand it well enough to see if there's another approach that would work. It might be worth checking with the tracing devs for any obvious workarounds as a third option.

joshka:
I suggest discussion continue at https://github.com/hyperium/hyper/issues/3870

arpadav:
@joshka I completely agree.

It's an extremely unfortunate sequence of events that:

* `reqwest` (and other dependents of `hyper-util`) use the client / legacy client features in `Cargo.toml`
* both client/legacy client depend on `dep:tracing` feature
* legacy client has _correct_ `tracing` API usage in DNS resolution future
* obscure `tracing` edge-case causes panic
* as a result, `hyper-util` gets the short-end of the stick and reverts a change to prevent panicking, despite doing nothing wrong

That's why in #179 I say it can be once again reverted, once `tracing` contributors/maintainers track down the source of this issue. 

Because of this, I'm motivated to make a simple test/dummy repo showing the panic vs the expected behavior. Just spend the last 30 min making it here: https://github.com/arpadav/tracing-span-bug

And then I / others can iterate on the `tracing` version to see when it starts to work again. Reason being, I am not crazy invested in this and wouldn't be surprised if `tracing` fixes it without me being aware, so having that repo is good sanity check.

PR #136, opened 2024-07-10
feat(server): implement Clone for GracefulShutdown

tottoto:
Implements Clone for GracefulShutdown. This is needed to use GracefulShutdown in other functions.


seanmonstar:
> in other functions.

Not that I'm against adding this, but know the use cases would help.

There's a part of me that worries if having multiple clones could be confusing when `shutdown()` consumes self...

tottoto:
https://github.com/hyperium/tonic/pull/1788 is a use case of this change.

seanmonstar:
I'm certainly sympathetic to the use case. I've bumped into a couple times, too.

But something always made me worry. I finally spent some time understanding my concern, and wrote up a different solution: #182

PR #142, opened 2024-08-07
feat(http): Add conditional compilation for `interface` field in Config struct.

0x676e67:
Ignore dead code warnings for interface fields on unsupported platforms


seanmonstar (approved):
Thanks! :)

PR #144, opened 2024-08-13
chore(dependencies): bump tower to v0.5

paolobarbolini:
Upgrades to the recently released [tower v0.5](https://github.com/tower-rs/tower/releases/tag/tower-0.5.0)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ socket2 = { version = "0.5", optional = true, features = ["all"] }
 tracing = { version = "0.1", default-features = false, features = ["std"], optional = true }
 tokio = { version = "1", optional = true, default-features = false  }
 tower-service ={ version = "0.3", optional = true }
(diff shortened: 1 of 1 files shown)


paolobarbolini:
Weird. `tower` is supposed to be Rust 1.63 but it's 1.64 :thinking:. Looking into it

jplatte:
Not sure whether it's a good idea, but hyper-util could also easily vendor `Oneshot` to keep its current MSRV.

jplatte:
@seanmonstar What's your preferred solution here? Worth noting that https://github.com/tower-rs/tower/pull/792 would relax tower's MSRV back to 1.63, but at the cost of reverting a previous change that was made to ease maintenance.

seanmonstar:
Hm, I can just vendor `Oneshot`, that likely is the easiest for all the projects, and it's code is very small.

seanmonstar:
I ended up vendoring Oneshot, and no longer depend on tower, in #151.

PR #148, opened 2024-09-14
feat: use non-zero ports resolved by dns resolvers

Nuhvi:
Partly solves https://github.com/seanmonstar/reqwest/issues/2413

### DIFF
--- a/src/client/legacy/connect/http.rs
+++ b/src/client/legacy/connect/http.rs
@@ -433,7 +433,12 @@ where
                 .map_err(ConnectError::dns)?;
             let addrs = addrs
                 .map(|mut addr| {
-                    addr.set_port(port);
+                    // Respect explicit ports in the URI,
+                    // and non `0` ports resolved from a custom dns resolver.
+                    if dst.port().is_some() || addr.port() == 0 {
+                        addr.set_port(port)
+                    };
+
                     addr
                 })
                 .collect();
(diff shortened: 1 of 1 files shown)


src/client/legacy/connect/http.rs
```diff
@@ -433,7 +433,12 @@ where
                 .map_err(ConnectError::dns)?;
             let addrs = addrs
                 .map(|mut addr| {
-                    addr.set_port(port);
+                    // Respect explicit ports in the URI,
+                    // and non `0` ports resolved from a custom dns resolver.
+                    if dst.port().is_some() || addr.port() == 0 {
```
seanmonstar, line 438: Because of the combination of logical conditionals, I think it'd be wise to extract this little bit of code into a function that we can write a few unit tests for inside this file, down at the bottom. Not tests going through the whole resolution step, just once an address is resolved, the combination of port missing, port explicitly in `dst`, and default scheme port. What do you think?

Nuhvi: Yes sure makes sense, I tested few combinations manually, and considered writing tests but the whole process was daunting to add tests too, testing a function for just this would be easy, will do.

Nuhvi:
Not sure how can I fix this CI failiure.

seanmonstar:
I've fixed it in #149.

seanmonstar (approved):
Nice, thanks for the tests, this is excellent <3

PR #152, opened 2024-10-01
Add a new config to set TCP_USER_TIMEOUT

guillaumebort:
Setting TCP keepalive is not enough to prevent dead TCP connections.

See https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/

I'm adding this to the "legacy" HTTP client because my goal is to add the setting to Tonic which is still using the `HttpConnector`.


seanmonstar (approved):
Looks good to me! Thanks :)

PR #153, opened 2024-10-01
refactor: allow resolve span to be disabled

seanmonstar:
Because of what's noted in https://github.com/seanmonstar/reqwest/issues/2435

### DIFF
--- a/src/client/legacy/connect/dns.rs
+++ b/src/client/legacy/connect/dns.rs
@@ -31,7 +31,7 @@ use std::{fmt, io, vec};
 
 use tokio::task::JoinHandle;
 use tower_service::Service;
-use tracing::{debug, debug_span};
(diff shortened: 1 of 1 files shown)


GeorgeHahn:
This also causes a panic when `tracing_subscriber::Registry` is in use (https://github.com/tokio-rs/tracing/issues/1656). I was able to work around it by downgrading to `hyper-util` 0.1.7.

seanmonstar:
@GeorgeHahn the change in this PR causes a panic?

GeorgeHahn:
Sorry, I should have worded that differently. This PR will fix a panic. The `or_current` call brings the current span into the spawned task and tracing will panic if that span has been closed by the time the task completes. This panic is a sharp edge in tracing's `Registry`.

PR #159, opened 2024-12-05
chore: remove legacy::client workaround for fixed race condition in mpsc channel

jlizen:
Doing a bit of cleanup, this race condition was closed back in 2021.

Ref: https://github.com/rust-lang/futures-rs/issues/909


seanmonstar (approved):
Thanks for cleanup!

PR #160, opened 2024-12-28
fix: Align the cfg compilation with the `socket2`

linyihai:
This shoud fixe https://github.com/hyperium/hyper/issues/3817

### DIFF
--- a/src/client/legacy/connect/http.rs
+++ b/src/client/legacy/connect/http.rs
@@ -111,45 +111,89 @@ impl TcpKeepaliveConfig {
         }
     }
 
-    #[cfg(not(any(
-        target_os = "aix",
-        target_os = "openbsd",
-        target_os = "redox",
-        target_os = "solaris"
-    )))]
+    #[cfg(
+        // See https://docs.rs/socket2/0.5.8/src/socket2/lib.rs.html#511-525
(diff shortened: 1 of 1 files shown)


seanmonstar:
I don't think it can be done with the cfg macro, since that still compiles to just an `if` expression, the code is still compiled and those methods don't exist. The cfg attribute is what is needed to prevent compiling methods calls that don't exist on that target.

linyihai:
Thank you.  The cfg attribute has been used.

dswij:
Have we checked that this fixes the issue?

linyihai:
> Have we checked that this fixes the issue?

Hmm, I currently have no `Haiku OS` to test.

vmlemon:
> > Have we checked that this fixes the issue?
> 
> Hmm, I currently have no `Haiku OS` to test.

Kinda limited, in terms of integration testing that I can do, at the moment (only have remote access to a machine,  via an Android VNC client, where I am, right now), but I was able to successfully compile this, in a Haiku VM, under QEMU, on MacOS:

Screenshot_20241231_172003_RVNC Viewer

Thanks, for working on this. :)

seanmonstar (approved):

PR #162, opened 2025-01-16
Expose legacy error kinds

richarddd:
Sometimes we might want to retry a request if for instance we have a partial write/read and server has dropped connection.
This PR allows us to match on the error kinds to determine further action.

### DIFF
--- a/src/client/legacy/client.rs
+++ b/src/client/legacy/client.rs
@@ -54,14 +54,14 @@ struct Config {
 
 /// Client errors
 pub struct Error {
-    kind: ErrorKind,
+    pub kind: ErrorKind,
     source: Option<Box<dyn StdError + Send + Sync>>,
     #[cfg(any(feature = "http1", feature = "http2"))]
     connect_info: Option<Connected>,
 }
 
 #[derive(Debug)]
-enum ErrorKind {
+pub enum ErrorKind {
     Canceled,
     ChannelClosed,
     Connect,


seanmonstar:
Thanks for the PR!

I appreciate the desire to be able to handle certain error conditions (I really do!), I also am hesitant to expose the enum like so (see some thoughts in this [write-up](https://seanmonstar.com/blog/pattern-matching-and-backwards-compatibility/)).

richarddd:
While I do agree with some points in your post, and its not a super straight forward solution, I don't fully agree that providing methods is the right way. In this case, it's about handling error scenarios, where a user is very likely to handle specific error scenarios, meaning that the match would probably have a `_ =>` arm. 
Naturally this assumption doesn't apply for all use cases or implementors but i'd argue it's better than having no way other than an hacky workaround to get the debug representation to determine the error cause :)
```rust
let error = format!("{:?}", err);
let is_send_request_error = error.ends_with("(SendRequest)");
```

vlovich:
If I'm reading it correctly, it sounds like like what @seanmonstar is actually saying is that `#[non_exhaustive]` just needs to be an annotation on the ErrorKind?

seanmonstar:
No, I don't mean just adding `non_exhaustive`. I'd rather not make the enum public, but I'm welcome to proposals on forwards-compatible ways to expose error information.

PR #164, opened 2025-01-24
builder: expose methods for determining if HTTP/1.1 or HTTP/2 support are enabled

tobz:
## Context

When utilizing TLS, a user must know if they have their server configured to handle HTTP/1.1 and/or HTTP/2 to properly determine which protocols can be negotiated over ALPN.

This PR adds two new methods to `hyper_util::server::conn::auto::Builder`: `is_http1_available` and `is_http2_available`, which tie back to the inner version value used to determine how to serve a connection. In exposing these, users can now programmatically build the ALPN protocol list rather than having to hardcode it, potentially ending up with a mismatch between the hardcoded list and the actual feature flags enabled for `hyper`/`hyper-util`.

### DIFF
--- a/src/server/conn/auto/mod.rs
+++ b/src/server/conn/auto/mod.rs
@@ -139,7 +139,7 @@ impl<E> Builder<E> {
             #[cfg(feature = "http2")]
             Some(Version::H2) => false,
             #[cfg(any(feature = "http1", feature = "http2"))]
-            _ => true
+            _ => true,
         }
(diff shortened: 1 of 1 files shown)


seanmonstar:
Seems like a fine addition, wanna run fmt?

tobz:
Doh! Should hopefully be all set now.

seanmonstar (approved):

PR #165, opened 2025-01-26
fix(client): Fix `HTTP/2` upgrade `WebSocket`

0x676e67:
tracking: https://github.com/hyperium/hyper/discussions/3840

This is the simplest fix, I don't know if there is something else I'm missing

### DIFF
--- a/src/client/legacy/client.rs
+++ b/src/client/legacy/client.rs
@@ -316,7 +316,7 @@ where
             } else {
                 origin_form(req.uri_mut());
             }
-        } else if req.method() == Method::CONNECT {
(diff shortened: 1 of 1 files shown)


seanmonstar:
Wow, nice detective work! Would you be up to adding a legacy test after `fn client_upgrade()` that checks the same thing but over HTTP/2? I wouldn't want to break this in a refactor.

If you don't have time, I can get to it eventually.

0x676e67:
OK, I'll do this.

0x676e67:
I’ve added a new HTTP/2 upgrade test. If you need anything else, feel free to let me know.

seanmonstar (approved):
Perfect, thanks!

PR #171, opened 2025-03-04
Add a proxy Matcher

seanmonstar:
cc https://github.com/hyperium/hyper/issues/3850

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,13 +18,16 @@ features = ["full"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


seanmonstar:
I think the `Builder` type needs more methods, and the ability to load from macOS and Windows system proxies (another optional cargo feature). The way I've been testing it can do enough is to see if I can replace the types in reqwest. Not the exact types, mind you. `reqwest::Proxy` will still be its own type, but it could use this `Matcher` internally.

seanmonstar:
Tests added, docs updated, and I have a local branch of reqwest using this. I think it's ready.

PR #172, opened 2025-03-06
feat(client): Add system proxy support for macOS

andrzejSulkowski:
cc https://github.com/hyperium/hyper/issues/3850
continued work on PR Draft #171

Because the "system-proxies" introduced quite a few lines of code, I decided to split the matcher.rs into multiple files (`builder.rs`, `matcher.rs`, `no_proxy.rs`, `utils.rs`)

- [x] Introduced setter methods for the Builder
- [x] MacOS System Proxies
~~- [ ] Windows System Proxies~~

Windows System Proxies should be implemented by someone who is using Windows or has a VM to impl. and test the code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,13 +18,16 @@ features = ["full"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
-hyper = "1.4.0"
+base64 = { version = "0.22", optional = true }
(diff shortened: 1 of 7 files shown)


seanmonstar:
Didn't mean for this to auto-close. I think it just did because the branch this was aimed at was merged.

If you want to update it and point at master, I'd be happy to merge!

PR #173, opened 2025-03-09
feat(client): add `proxy::Socks` legacy util

JPDye:
cc hyperium/hyper#3851
builds atop hyperium/hyper-util#140

Adds `Socks` connector. Didn't pull existing crates as it's a simple enough protocol. Behaviour is there, clean up still necessary for error handling, etc...

`connect()` function will be refactored to enum-based state machine to make optimistic sending of messages cleaner. Optimistic sending isn't in the [RFC](https://datatracker.ietf.org/doc/html/rfc1928) but most existing server implementations can handle it from my testing. Worth adding for RTT reduction; will document to make clear it might not work.

### DIFF
--- a/src/client/legacy/connect/mod.rs
+++ b/src/client/legacy/connect/mod.rs
@@ -74,6 +74,8 @@ pub mod dns;
 #[cfg(feature = "tokio")]
 mod http;
 
+pub mod proxy;
+
 pub(crate) mod capture;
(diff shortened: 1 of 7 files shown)


seanmonstar:
Awesome work in here! I just now after looking through it realized it is a PR to my draft PR, so I probably need to fix that up.

PR #176, opened 2025-03-26
enable `HttpConnector::interface` on macOS and Solarish systems

hawkw:
The `client::legacy::connect::HttpConnector` type has a method
`set_interface` that configures the connector to bind TCP connections
only on the provided interface name. Presently, this is enabled only on
Linux, Android, and Fuchsia, as it uses the Linux-only socket option
[`SO_BINDTODEVICE`].

[On macOS] (and other Apple operating systems such as iOS, watchOS,
etc), [and Solaris] and Solaris-like systems like illumos, the Solaris
socket option [`IP_BOUND_IF`] provides similar functionality to
`SO_BINDTODEVICE`, as it also restricts connections to a particular
network itnerface. Unlike the Linux socket option, `IP_BOUND_IF` takes
the numeric _index_ of the interface, so in order to accept a string
interface name similarly to the Linux implementation, we must use
[`if_nametoindex(3socket)`] from libc to determine the index
corresponding to the interface name. This is done every time we open a
connection, to avoid TOCTOU issues in case interface indices or names
change.

Note that this PR currently requires an unreleased change in `socket2`
(rust-lang/socket2#566) to enable `IP_BOUND_IF` on Solarish operating
systems. We will need to remove the Cargo patch for `socket2` once that
change is released, prior to merging this PR.

[`SO_BINDTODEVICE`]: https://linux.die.net/man/7/socket
[On macOS]: https://developer.apple.com/forums/thread/714370
[and Solaris]:
    https://docs.oracle.com/cd/E86824_01/html/E54777/ip-7p.html
[`if_nametoindex(3socket)`]:
    https://smartos.org/man/3socket/if_nametoindex

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ http-body = "1.0.0"
 bytes = "1.7.1"
 pin-project-lite = "0.2.4"
 futures-channel = { version = "0.3", optional = true }
-socket2 = { version = "0.5", optional = true, features = ["all"] }
+socket2 = { version = "0.5.9", optional = true, features = ["all"] }
 tracing = { version = "0.1", default-features = false, features = ["std"], optional = true }
 tokio = { version = "1", optional = true, default-features = false  }
 tower-service = { version = "0.3", optional = true }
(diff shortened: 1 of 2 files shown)


seanmonstar:
LGTM, thanks! Will just wait till that socket2 release :)

hawkw:
> LGTM, thanks! Will just wait till that socket2 release :)

Yup, no rush --- just wanted to get this ready to go when that's out. :)

hawkw:
Haha, I was just about to leave a comment mentioning that the `socket2` release had been published --- impressive reaction time, @seanmonstar! :)

PR #178, opened 2025-04-02
fix(server): Enforce `serve_connection` result usage

Sol-Ell:
- Mark `serve_connection` and `serve_connection_with_upgrades` as `must_use` to prevent ignored results.  
- Mark `serve_connection_with_upgrades` as `must_use` for consistency.  
- Helps avoid silent issues when forgetting to `.await` the connection.  

Not a major issue, but spent 15 minutes debugging due to a missing `.await`.

### DIFF
--- a/src/server/conn/auto/mod.rs
+++ b/src/server/conn/auto/mod.rs
@@ -495,7 +495,8 @@ where
 }
 
 pin_project! {
-    /// An future binding a connection with a Service with Upgrade support.
+    /// An upgradable [`Connection`], returned by
+    /// [`Builder::serve_upgradable_connection`](struct.Builder.html#method.serve_connection_with_upgrades).
     ///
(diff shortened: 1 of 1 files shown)


seanmonstar (changes requested):
Oh, thanks for PR! Yea, we generally try to provide warnings for people like this. I assumed that `must_use` on `std::future::Future` would make it so we didn't need to do add it manually anymore, but it seems that's only for `impl Future` and `Box<dyn Future>`.

One request: can they be added to the types themselves, not to the functions, with a note? That's how we've done it for other futures: https://github.com/hyperium/hyper/blob/c449528a33d266a8ca1210baca11e5d649ca6c27/src/server/conn/http1.rs#L38

Sol-Ell:
Thanks for pointing this out. I'll update everything accordingly.

seanmonstar (approved):
Lovely, thank you!

PR #179, opened 2025-04-02
fix(tracing): revert #134 - tracing span removal in legacy DNS `GaiResolver`

arpadav:
`tracing` has an unresolved bug when attempting to fetch a span that _should_ exist. Tracking: https://github.com/tokio-rs/tracing/issues/3223#issuecomment-2771038993

#134 introduces this bug. `hyper` contributors aren't at fault — this is a `tracing` issue.

As a result, crates that depend on `hyper`'s legacy client (e.g. `reqwest`) will immediately panic when a request is sent under specific circumstances configured by `tracing` spans. (The tracked comment shows a pseudocode example). This slightly more accurate `tracing` span is not worth the panics. 

In theory, this PR can be reverted once `tracing` resolves this issue, but its up to the maintainers whether or not it should be prioritized.

### DIFF
--- a/src/client/legacy/connect/dns.rs
+++ b/src/client/legacy/connect/dns.rs
@@ -30,7 +30,6 @@ use std::{fmt, io, vec};
 
 use tokio::task::JoinHandle;
 use tower_service::Service;
-use tracing::debug_span;
 
 pub(super) use self::sealed::Resolve;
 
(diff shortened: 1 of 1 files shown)


joshka:
The impact of reverting this is that DNS requests will no longer be under the same span that tracks the request. This may affect tracing and metrics on any places where this info is tracked. That said, this downside is less problematic than the alternative which causes panics, so I'm for this being merged

joshka (approved):

joshka:
Captured the overall details of the span in DNS problem in https://github.com/hyperium/hyper/issues/3870

seanmonstar (approved):

PR #182, opened 2025-04-14
feat(server): add graceful::Watcher type

seanmonstar:
This allows creating an owned type that can be passed to another task before `watch`ing the connection.

I also included some of the "why" in code comments, why it cannot just be `Clone`.

Supercedes #136


tottoto:
I agree that this is more appropriate than implementing clone.

tottoto (approved):

PR #184, opened 2025-04-18
fix(client): race in connection errors propagation

dare3path:
Fix a race condition in the legacy HTTP client's connection setup where connection errors (e.g., TLS failures, unexpected server responses) were discarded, resulting in vague ChannelClosed errors.

https://github.com/seanmonstar/reqwest/issues/2649

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,8 @@ http-body-util = "0.1.0"
 tokio = { version = "1", features = ["macros", "test-util", "signal"] }
 tokio-test = "0.4"
 pretty_env_logger = "0.5"
+tracing-subscriber = { version = "0.3", features = ["env-filter", "fmt"] }
+ctor = "0.2"
 
 [target.'cfg(any(target_os = "linux", target_os = "macos"))'.dev-dependencies]
 pnet_datalink = "0.35.0"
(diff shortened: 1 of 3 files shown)


dare3path:
Apologies, I forgot to mention that I accidentally pushed a commit to this branch not realizing it was a branch that's part of a PR, then I undid it by force-pushing the branch such that it did not contain the mistakenly pushed commit, to preserve the initial state. Sorry. It's all good now (since 2 days ago) and thanks for re-running the CI on it.

seanmonstar:
Thanks for fixing this, I now see how this fits together, makes sense!

I wonder, can you think of a way we could deterministically test this? We can control a mock IO pretty strictly with [`tokio_test::io`](https://docs.rs/tokio-test/latest/tokio_test/io/index.html). It doesn't need to have tests here for me to merge, but it might be a helpful follow up if you're interested, to help prevent relapse as we refactor the legacy client (into a new one, eventually).

dare3path:
I'll look into it, though I should mention that after something like 2 days or more with grok3 I failed to make tests here in `hyper-util`(without using `reqwest`), however since I don't know much, I didn't know about that mock IO that you mentioned, maybe with this new info grok3&i could come up with some reliable tests. I'll try and let you know what comes of it. If in the meantime you decide otherwise, please let me know. I'm on it until then or until something comes of it.

dare3path:
success

dare3path:
> Could you remove the part about using ctor? I don't think we need that. We don't need the init tracing part either.

done, and rebased on current master (8805922b8230160f08ca5f05ae1003c07f9260f4)

seanmonstar (approved):

PR #186, opened 2025-04-20
docs(client): Include .pool_timer() in the Client builder example

sulami:
The example (/doctest) uses a pool idle timeout, which according to the docs doesn't actually do anything without a pool timer.

Including a timer makes this example less misleading.


dswij (approved):

seanmonstar (approved):
Good point!

PR #191, opened 2025-05-08
Add Hermit support

mkroening:
This PR adds support for the Hermit OS. For networking, we try to adhere to POSIX, so we should be similar to Unix.

### DIFF
--- a/src/client/legacy/connect/http.rs
+++ b/src/client/legacy/connect/http.rs
@@ -911,13 +911,16 @@ fn connect(
     )
     .map_err(ConnectError::m("tcp bind local error"))?;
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
Hey there, thanks for the PR! It seems CI notice a formatting issue.

mkroening:
Thanks for the review!

After going through a few other crates and noticing increasing duplication due to Hermit getting ever closer to `std::os::unix`, we are currently reconsidering marking Hermit as Unix-like in rustc, which would make this PR obsolete.

I will close this PR and reopen it if we do not go forward with the plan to officially mark Hermit as Unix.

PR #192, opened 2025-05-17
refactor: limit dependency futures-util to tests and client-legacy

hanna-kruppe:
Enable non-trivial applications based on hyper + tokio without pulling in this heavyweight dependency. Keep MSRV 1.63 by using `ready!` from `futures_core` and polyfilling `poll_fn`.

Don't remove futures-util from the `client-legacy` code, because this would require re-implementing several non-trivial combinators.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,8 @@ rustdoc-args = ["--cfg", "docsrs"]
 base64 = { version = "0.22", optional = true }
 bytes = "1.7.1"
 futures-channel = { version = "0.3", optional = true }
-futures-util = { version = "0.3.16", default-features = false }
+futures-core = { version = "0.3" }
+futures-util = { version = "0.3.16", default-features = false, optional = true }
 http = "1.0"
 http-body = "1.0.0"
 hyper = "1.6.0"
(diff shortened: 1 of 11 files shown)


.github/workflows/CI.yml
```diff
@@ -60,7 +60,7 @@ jobs:
     needs: [style]
     strategy:
       matrix:
-        rust: [ 1.63 ] # keep in sync with 'rust-version' in Cargo.toml
+        rust: [ 1.64 ] # keep in sync with 'rust-version' in Cargo.toml
```
hanna-kruppe, line 63: It's possible to keep MSRV the same way I did in the accompanying hyper PR (re-implement `poll_fn`, import `ready!` from `futures_core`) if that's preferred.

seanmonstar: Just checked, Debian stable is still on 1.63. While I don't explicitly document that we track Debian, it is one of the gauges I use to determine if it's worth increasing MSRV.

hanna-kruppe: Rebased and updated to keep MSRV as-is.

seanmonstar (approved):

PR #198, opened 2025-05-28
Update to `windows-registry` 0.5

kennykerr:
This version of [windows-registry](https://crates.io/crates/windows-registry) is [largely the same](https://github.com/microsoft/windows-rs/releases/tag/64) but greatly reduces your dependencies as it uses the tiny [windows-link](https://crates.io/crates/windows-link) crate for linking Windows imports rather than the older [windows-targets](https://crates.io/crates/windows-targets) crate that pulls down a bunch of large import libs.


seanmonstar (approved):
Thanks!

PR #207, opened 2025-06-12
feat(builder): support header case methods

AndreyErmilov:
Hey, it’d be great to have the ability to use `title_case_headers` and `preserve_header_case` with `auto::Builder`. Could you take a look at this PR?


seanmonstar (approved):
Looks good, thanks!

PR #209, opened 2025-07-05
Update to socket2 v0.6

Thomasdezeeuw:


seanmonstar:
While our policy allows for upgrading, I try pretty hard to support 1.63 for Debian.

seanmonstar:
Yea if I can pin a dependency for CI, I find that valid: users can choose to pin as well, or use a newer version, they get the choice.

tottoto:
Since the test CI is successful with just the version update, it seems that the current usage is not affected by the breaking change. If that is the case, it might be possible to keep the MSRV by allowing both 0.5.9 and 0.6 versions.

PR #218, opened 2025-07-19
Move tokio/net feature to client to make TokioIo can used in wasm

tiannian:
I was trying to port an http service using hyper-util to cloudflare worker but meet a dependency issue with mio. It caused by `TokioIo` need `tokio` feature, but this feature enable `tokio/net`. However, this was not necessary when using TokioIo alone, so I thought I could move this feature to the `client`.


seanmonstar (changes requested):
Thanks, this is great!

Could you revert most of the formatting changes? (The ones for `client` and `client-legacy` are fine)

seanmonstar (approved):
LGTM, thanks!

PR #226, opened 2025-09-15
feat(pool): add a Singleton pool type

seanmonstar:
This creates a `Singleton` pool service which wraps an inner make-service, and queues up all calls while waiting for the one instance to be created, and then hands out clones. This pattern fits HTTP/2 well.

The current implementation mirrors much of how the legacy pool handles connecting with HTTP/2 prior knowledge.

Closes https://github.com/hyperium/hyper/issues/3953

### DIFF
--- a/src/client/pool/singleton.rs
+++ b/src/client/pool/singleton.rs
@@ -1,0 +1,493 @@
+//! Singleton pools
+//!
+//! This ensures that only one active connection is made.
+//!
+//! The singleton pool wraps a `MakeService<T, Req>` so that it only produces a
+//! single `Service<Req>`. It bundles all concurrent calls to it, so that only
+//! one connection is made. All calls to the singleton will return a clone of
+//! the inner service once established.
+//!
+//! This fits the HTTP/2 case well.
+//!
+//! ## Example
+//!
+//! ```rust,ignore
+//! let mut pool = Singleton::new(some_make_svc);
+//!
+//! let svc1 = pool.call(some_dst).await?;
+//!
+//! let svc2 = pool.call(some_dst).await?;
+//! // svc1 == svc2
+//! ```
+
+use std::sync::{Arc, Mutex};
+use std::task::{self, Poll};
+
+use tokio::sync::oneshot;
+use tower_service::Service;
+
+use self::internal::{DitchGuard, SingletonError, SingletonFuture, State};
+
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
+#[cfg(docsrs)]
+pub use self::internal::Singled;
+
+/// A singleton pool over an inner service.
+///
+/// The singleton wraps an inner service maker, bundling all calls to ensure
+/// only one service is created. Once made, it returns clones of the made
+/// service.
+#[derive(Debug)]
+pub struct Singleton<M, Dst>
+where
+    M: Service<Dst>,
+{
+    mk_svc: M,
+    state: Arc<Mutex<State<M::Response>>>,
+}
+
+impl<M, Target> Singleton<M, Target>
+where
+    M: Service<Target>,
+    M::Response: Clone,
+{
+    /// Create a new singleton pool over an inner make service.
+    pub fn new(mk_svc: M) -> Self {
+        Singleton {
+            mk_svc,
+            state: Arc::new(Mutex::new(State::Empty)),
+        }
+    }
+
+    // pub fn clear? cancel?
+
+    /// Retains the inner made service if specified by the predicate.
+    pub fn retain<F>(&mut self, mut predicate: F)
+    where
+        F: FnMut(&mut M::Response) -> bool,
+    {
+        let mut locked = self.state.lock().unwrap();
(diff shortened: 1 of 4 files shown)


src/client/pool/singleton.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    // XXX: pub because of the enum SingletonFuture
+    pub struct DitchGuard<S>(pub(super) Weak<Mutex<State<S>>>);
+
+    impl<F, S, E> Future for SingletonFuture<F, S>
+    where
+        F: Future<Output = Result<S, E>>,
+        E: Into<BoxError>,
+        S: Clone,
+    {
```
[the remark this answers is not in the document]
cratelyn: on further reading, i believe i see why each of these won't work: one notable bit is that multiple calls to the `MakeService<T>` may accrue while we establish the connection.

src/client/pool/singleton.rs
```diff
@@ -0,0 +1,333 @@
...
+    M: Service<Dst>,
+{
+    mk_svc: M,
+    state: Arc<Mutex<State<M::Response>>>,
+}
+
+#[derive(Debug)]
+enum State<S> {
+    Empty,
+    Making(Vec<oneshot::Sender<S>>),
```
cratelyn, line 37: this should be bounded in some manner! if the `SingletonFuture` gets stuck whilst establishing a connection, we could end up with an unbounded number of waiters here.

it'd be nice if configuring that was optional (_e.g. with a builder-style method?_) with a relatively high default value.

ideally, we also introduce some lightweight test coverage to ensure that only N waiters can enter the queue once we've begun "making" the inner service.

src/client/pool/singleton.rs
```diff
@@ -0,0 +1,333 @@
...
+                                singleton.0 = Weak::new();
+                                match std::mem::replace(&mut *locked, State::Made(svc.clone())) {
+                                    State::Making(waiters) => {
+                                        for tx in waiters {
+                                            let _ = tx.send(svc.clone());
+                                        }
+                                    }
+                                    State::Empty | State::Made(_) => {
+                                        // shouldn't happen!
+                                    }
```
cratelyn, line 170: it'd be nice if this emitted a `tracing::warn!` should we ever hit this, ideally behind a feature gate.

src/client/pool/singleton.rs
```diff
@@ -0,0 +1,333 @@
...
+                }
+            }
+            State::Making(ref mut waiters) => {
+                let (tx, rx) = oneshot::channel();
+                waiters.push(tx);
+                SingletonFuture::Waiting { rx }
+            }
+            State::Made(ref svc) => SingletonFuture::Made {
+                svc: Some(svc.clone()),
+            },
```
cratelyn, line 96: something we may wish to add as a follow-on to this would be a similar "singleton" pool that is backed by an `Arc<T>`. upon my initial read, i was _slightly_ surprised we weren't `Arc<T>`'ing the generated `Service<T>` to already.

as is, this calls `mk_svc.call()` once, but does result in multiple distinct copies of the `M::Response` service. an `Arc<T>`-backed singleton pool would be useful for services that are expensive to `Clone`.

i don't believe any of this blocks this specific change, but i'm curious to ask where we think it might reasonably live. ideally we could rely on a lot of this infrastructure, since 99% of it will be the same! 😸

src/client/pool/singleton.rs
```diff
@@ -0,0 +1,333 @@
...
+
+use self::internal::{DitchGuard, SingletonError, SingletonFuture};
+
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
+/// A singleton pool over an inner service.
+///
+/// The singleton wraps an inner service maker, bundling all calls to ensure
+/// only one service is created. Once made, it returns clones of the made
+/// service.
```
cratelyn, line 24: i'd suggest that we clarify this slightly, and indicate that we _lazily_ instantiate the inner service.

i could imagine an alternative "eager" singleton connection pool that invokes the `MakeService` immediately, for example.

src/client/pool/singleton.rs
seanmonstar: I thought about if it needed bounding, but figured bounding it also something that can happen with an additional layer, just slotting in `ConcurrenyLimit` layer, for instance. In that case, it feels like the appropriate thing for this layer is to act just like a vec, where the bound is just memory.

src/client/pool/singleton.rs
cratelyn: oh, that's a very nice idea! `debug_assert!(false, "this should not happen")` would work well.

src/client/pool/singleton.rs
cratelyn: that makes sense to me. thank you, i will mark this thread as resolved!

cratelyn (approved):
modulo the note(s) on documentation comments, i think this looks great. i am excited to see this work move forward! nice work 🚀

PR #278, opened 2026-03-27
feat(server,http1,http2,upgrade): remove body static lifetime constraint

Voldemat:
Before this change, hyper expects body trait implementation to have 'static lifetime, which prevents use cases where one might have StreamBody that references certain state that outlives hyper connection object. This change adds lifetime parameter 'body to varies hyper objects that replaces 'static.


cratelyn:
@Catwoman08 noted that this is not planned in https://github.com/hyperium/hyper/issues/4096#issuecomment-5319021996. closing this as well.

PR #279, opened 2026-03-31
fix: fix is_cancelled checked error

dentiny:
Reading from the [code](https://github.com/hyperium/hyper-util/blob/8ae9e8b1e338a5b7d35155a0dc3708cfd94bcae2/src/client/legacy/pool.rs#L620), `CheckoutNoLongerWanted` indicates cancellation.

### DIFF
--- a/src/client/legacy/pool.rs
+++ b/src/client/legacy/pool.rs
@@ -608,7 +608,7 @@ pub enum Error {
 
 impl Error {
     pub(super) fn is_canceled(&self) -> bool {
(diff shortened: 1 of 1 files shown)


cratelyn:
thank you for opening a pull request, and my apologies for the amount of time it took to get this reviewed. unfortunately, i don't think we can safely merge this.

you might be right that this is a bug in the legacy client code, but because it exists in part to facilitate migration from previous releases (e.g. hyper v0.14) i don't believe that we should make behavioral changes to this code.

PR #280, opened 2026-04-23
proxy: honor Windows proxy bypass macros and loopback defaults

chitao1234:
Parse Windows `ProxyOverride` entries into explicit no-proxy rules while supporting the special `<local>` and `<-loopback>` semantics.

By default, preserve the system loopback bypass behavior for `localhost`, `loopback`, `127.0.0.1`, and `::1`, and add tests covering local-name and loopback matching.

Source:
<https://learn.microsoft.com/en-us/windows/win32/wininet/enabling-internet-functionality>
<https://learn.microsoft.com/en-us/windows/win32/api/wininet/nf-wininet-internetopena>

### DIFF
--- a/src/client/proxy/matcher.rs
+++ b/src/client/proxy/matcher.rs
@@ -13,7 +13,7 @@
 //! authentication to be used.
 
 use std::fmt;
-use std::net::IpAddr;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 
 use http::header::HeaderValue;
 use ipnet::IpNet;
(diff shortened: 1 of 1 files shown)


src/client/proxy/matcher.rs
```diff
@@ -66,6 +68,8 @@ enum Auth {
 struct NoProxy {
     ips: IpMatcher,
     domains: DomainMatcher,
+    local_names: bool,
+    loopback_hosts: bool,
```
seanmonstar: I'm curious, is there a reason to store additional booleans instead of just pushing `localhost` and the IPs to the `DomainMatcher` and `IpMatcher`?

chitao1234: I agree that the loopback flags can be folded, and I folded them into the windows proxy override parsing logic. However the local flags should be kept due to there not being an equivalent for it (it matches all domain names without a dot in the middle like `http://intranet/`).

PR #286, opened 2026-05-22
chore: improve hyper-util maintenance path

oab24413gmai:
Summary:
- Add unit tests for the Rewind type covering pre-buffered read paths, partial reads spanning the rewound buffer and inner reader, and empty-prebuf behavior to lock in correctness of the AsyncRead implementation
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/src/common/rewind.rs
+++ b/src/common/rewind.rs
@@ -85,53 +85,105 @@ where
     }
 }
 
-/*
 #[cfg(test)]
 mod tests {
     use super::Rewind;
     use bytes::Bytes;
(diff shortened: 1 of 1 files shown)


seanmonstar:
It looks like this was posted 4 times, so I assume we can close #287, #288, and #289.

sjcobb2022:
@seanmonstar this is definitely either just spam or someone has pointed a openclaw instance here or something.

PR #292, opened 2026-06-12
fix(client): cancel idle interval when pool empties

lrowe:
The IdleTask is started on first connection when pool_idle_timeout is set however it is not stopped when the pool is empty leaving the event loop spinning every pool_idle_timeout ms.

I noticed this in Deno (see: https://github.com/denoland/deno/issues/29444) and the fix seems fairly small. Do you think such a change would be worth considering? Happy to work this up with test cases if so.

### DIFF
--- a/src/client/legacy/pool.rs
+++ b/src/client/legacy/pool.rs
@@ -798,6 +798,11 @@ impl<T: Poolable + 'static, K: Key> IdleTask<T, K> {
                         if let Ok(mut inner) = inner.lock() {
                             trace!("idle interval checking for expired");
                             inner.clear_expired();
+                            if inner.idle.is_empty() {
+                                inner.idle_interval_ref = None;
+                                trace!("pool empty, canceling idle interval");
+                                return;
+                            }
                         }
                     }
 


Catwoman08:
This is tied to https://github.com/hyperium/hyper/issues/4100

seanmonstar:
Yea, I think this is worth fixing. And it should "just work", a new connection will see `None` and start a new timer task. Some sort of unit test would be good to add to the legacy pool module.

seanmonstar (changes requested):
Just needs a test, I think.

lrowe:
I've added an assert to test_pool_timer_removes_expired that `idle_interval_ref.is_none()`.

src/client/legacy/pool.rs
```diff
@@ -1021,19 +1011,21 @@ mod tests {
...
         );
+        assert!(pool.locked().idle_interval_ref.is_some());
 
         // Now wait passed the minimum interval more
         tokio::time::sleep(Duration::from_millis(70)).await;
         // Yield in case other task hasn't been able to run :shrug:
         tokio::task::yield_now().await;
 
         assert!(!pool.locked().idle.contains_key(&key));
+        assert!(pool.locked().idle_interval_ref.is_none());
```
seanmonstar, line 1022: Perhaps for robustness, we should test that a new item recreates it? We're implicitly depending on the code to create a timer if it is none, it'd be good to catch ourselves if we every moved that to the constructor or some other only-once place.

seanmonstar (approved):

PR #293, opened 2026-06-19
fix(client): parse Windows per-scheme proxy settings

### LINKED ISSUE #4067: Windows system proxy: per-scheme ProxyServer format (e.g. socks=127.0.0.1:1080) not parsed
### Version

hyper-util 0.1.20 (also present in 0.1.16, 0.1.19)

### Platform

Windows

### Summary

On Windows, Matcher::from_system() reads the registry value HKCU\Software\Microsoft\Windows\CurrentVersion\Internet Settings\ProxyServer and assigns its raw string verbatim to builder.http and builder.https. This only works for the
  single-proxy form. The per-scheme WinINET form (e.g. http=127.0.0.1:8080;https=127.0.0.1:8443;socks=127.0.0.1:1080) is not parsed, which results in malformed proxy URIs like http://socks=127.0.0.1 being produced downstream.

### Code Sample

```rust
Configure Windows system proxy with per-scheme entries, so the registry contains, for example:

ProxyServer = socks=127.0.0.1:1080

(This is what Windows writes when only a SOCKS proxy is configured via the Internet Settings UI in some setups, and the same format applies when multiple schemes are configured: http=...;https=...;socks=....)

Then build a client that uses system proxies (e.g. reqwest with the system-proxy feature, which calls hyper_util::client::proxy::matcher::Matcher::from_system()).
```

### Expected Behavior

- socks=host:port should be routed as a SOCKS proxy for all schemes (WinINET semantics: SOCKS applies to all).
- http=host:port should populate the HTTP proxy.
- https=host:port should populate the HTTPS proxy.
- Other tokens (ftp=..., etc.) should be ignored without breaking the rest.

### Actual Behavior

The full registry string (e.g. socks=127.0.0.1:1080) is assigned verbatim to both builder.http and builder.https. parse_env_uri() then parses it with http::Uri, finds no scheme, and defaults to http, producing an effective URI like

http://socks=127.0.0.1:1080 — not a usable proxy.

### Additional Context

_No response_

JoakoI98:
On Windows, `Matcher::from_system()` assigned the raw `ProxyServer` registry value to the HTTP/HTTPS proxies. That only worked for the single-proxy form (`host:port`)
The per-scheme WinINET form (`http=...;https=...;socks=...`) was left unparsed, producing malformed uris.

Closes https://github.com/hyperium/hyper/issues/4067.

This parses the per-scheme form before handing values to `parse_env_uri`:

- `http=`/`https=` populate their slots; the authority is passed through, so a scheme-less value defaults to an HTTP proxy (which tunnels HTTPS via CONNECT).
- `socks=` applies to all schemes via the `all` slot, normalized to `socks5://`.
- Unknown tokens (`ftp=`, etc.) and empty segments are ignored, and a value with no `=` keeps the previous single-proxy behavior.
- Environment variables keep priority; registry values only fill empty slots.

Unit tests for the parser are added under the `windows` module.

### DIFF
--- a/src/client/proxy/matcher.rs
+++ b/src/client/proxy/matcher.rs
@@ -662,6 +662,66 @@ mod mac {
 #[cfg(feature = "client-proxy-system")]
 #[cfg(windows)]
 mod win {
+
+    struct WindowsProxy {
+        http: Option<String>,
+        https: Option<String>,
+        all: Option<String>,
+    }
+
+    const HTTP_SCHEME: &str = "http";
+    const HTTPS_SCHEME: &str = "https";
+    const SOCKS_SCHEME: &str = "socks";
+
+    const SOCKS_WINDOWS_MAP_TO: &str = "socks5";
+
+    impl WindowsProxy {
+        fn uri_from_schema_and_authority(scheme: &str, authority: &str) -> String {
+            let safe_authority = authority
+                .split_once("://")
+                .map(|(_scheme, authority)| authority)
+                .unwrap_or(authority);
+            format!("{scheme}://{safe_authority}")
+        }
+
+        fn from_proxy_server_setting(proxy_server_setting: &str) -> Self {
+            let mut http = None;
+            let mut https = None;
+            let mut all = None;
+            for proxy in proxy_server_setting
+                .split(';')
+                .map(str::trim)
+                .filter(|p| !p.is_empty())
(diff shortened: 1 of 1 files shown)


dswij:
@JoakoI98 Thanks for the PR, can you help to resolve the conflicts?

PR #296, opened 2026-06-29
fix(pool): shared Singleton errors with all waiters

### LINKED ISSUE #4119: Composable pool: coalesced Singleton waiters are canceled when the maker bounces to the fallback leg
### Version

hyper-util:0.1.20

### Platform

macos aarch64

## Summary

In hyper-util 0.1.20's composable pool layers (`hyper_util::client::pool::{negotiate, singleton}`), concurrent same-authority requests through the eager `Negotiate` stack coalesce on the `Singleton` upgrade leg. When the maker discovers the connection is not h2 and returns `UseOther` to bounce to the H1 fallback, its error path resets the Singleton state and drops the coalesced waiters' oneshot senders. Each waiter resolves `Err(SingletonError)` with source `"singleton connection canceled"` instead of obtaining a connection. Exactly one caller (the maker) succeeds; all other concurrent callers to that authority fail spuriously.

## Mechanism

1. `Negotiate::call` starts in `Eager` state and polls the upgrade (`Singleton`) leg first.
2. Concurrent calls to clones of the same `Negotiate` (sharing `Arc<Mutex<State>>`) coalesce in the `Singleton`: the first becomes the maker (`State::Making`), the rest become waiters holding `oneshot::Receiver`s, with their senders stored in the `Making` waiter `Vec`.
3. The maker's inner future finds the slot empty (no h2 negotiation occurred) and returns `Err(UseOther)`.
4. In `SingletonFuture`, the maker's `Driving` arm hits the error branch (`singleton.rs:243-250`): it replaces the shared state with `State::Empty`, dropping the `State::Making(waiters)` value and thereby every waiter sender.
5. Each waiter's `Waiting` arm polls its `oneshot::Receiver`, which resolves `Err(Canceled)`, and returns `Err(SingletonError(Canceled.into()))` (`singleton.rs:255`).
6. The maker's `UseOther` propagates up to its own `Negotiate`, which bounces to the fallback leg and succeeds. The waiters have no equivalent path and are orphaned.

`DitchGuard::drop` resets state the same way when the maker future is dropped or cancelled before resolving (the path exercised by the existing `cancel_driver_cancels_all` test).

## Context

`singleton.rs` test `cancel_driver_cancels_all` currently asserts this cancel behavior and carries a `// TODO: this should be able to be improved with a cooperative baton refactor` comment, indicating the orphaning is acknowledged.

## Environment

- hyper-util 0.1.20 (crates.io release)
- Features: `client`, `client-pool`
- `tokio` `multi_thread` runtime (deterministic; does not surface on `current_thread`)

### Code Sample

```rust
## Reproduction

Minimal standalone crate: https://github.com/aajtodd/repro-hyper-util-singleton-cancel


cargo test -- --nocapture


Observed:


BUG REPRODUCED: 1/8 succeeded, 7/8 got SingletonError(Canceled)


The repro builds an eager `negotiate` stack with a `Singleton` upgrade leg, an identity fallback leg, and an `inspect` closure that always reports not-h2. It readies N clones of the negotiate service, calls all N synchronously, then polls the returned futures together on a `multi_thread` runtime.
```

### Expected Behavior

Expected: every coalesced caller obtains a connection — either the maker's established connection or its own via the fallback leg.

### Actual Behavior

Observed: the maker succeeds; all coalesced waiters resolve `Err(SingletonError)` with source `"singleton connection canceled"`.

Impact: any consumer issuing concurrent requests to the same authority through the composable pool with the eager negotiate stack sees most concurrent same-authority requests fail spuriously under load rather than reusing or falling back.

Additionally there is no typed error to key off of for this. 

### Additional Context

_No response_

seanmonstar:
Previously, when a bunch of calls to the singleton made a list of waiters, but the underlying service errored making the service, only the first caller would get the underlying error. All the rest of the waiters would just be told the operation was canceled. Which is true, but it made it harder for higher layers to know what to do.

Now, since they are all shared the make call, any error that happens is also shared to all the waiters. We do this by putting the error inside an `Arc<dyn std::error::Error>` internally, though that exact type isn't exposed. The source chain now properly will show a caller the underlying problem.

Closes https://github.com/hyperium/hyper/issues/4119

cc @aajtodd

### DIFF
--- a/src/client/pool/singleton.rs
+++ b/src/client/pool/singleton.rs
@@ -103,7 +103,7 @@ where
             return self
                 .mk_svc
                 .poll_ready(cx)
-                .map_err(|e| SingletonError(e.into()));
+                .map_err(|e| SingletonError::new(e.into()));
         }
         Poll::Ready(Ok(()))
     }
@@ -151,7 +151,7 @@ where
 mod internal {
     use std::future::Future;
     use std::pin::Pin;
-    use std::sync::{Mutex, Weak};
+    use std::sync::{Arc, Mutex, Weak};
     use std::task::{self, Poll, ready};
 
     use pin_project_lite::pin_project;
@@ -169,7 +169,7 @@ mod internal {
                 singleton: DitchGuard<S>,
             },
             Waiting {
-                rx: oneshot::Receiver<S>,
+                rx: oneshot::Receiver<Result<S, SharedError>>,
                 state: Weak<Mutex<State<S>>>,
             },
             Made {
@@ -183,7 +183,7 @@ mod internal {
     #[derive(Debug)]
     pub enum State<S> {
         Empty,
-        Making(Vec<oneshot::Sender<S>>),
+        Making(Vec<oneshot::Sender<Result<S, SharedError>>>),
         Made(S),
     }
 
@@ -227,7 +227,7 @@ mod internal {
                                 match std::mem::replace(&mut *locked, State::Made(svc.clone())) {
                                     State::Making(waiters) => {
                                         for tx in waiters {
-                                            let _ = tx.send(svc.clone());
+                                            let _ = tx.send(Ok(svc.clone()));
                                         }
                                     }
                                     State::Empty | State::Made(_) => {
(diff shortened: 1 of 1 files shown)


aajtodd:
I ran the #4119 repro against this branch: waiter outcome goes from 1/8 to 8/8 succeeding, deterministic across 5 runs.

If I'm understanding right, the mechanism for why the waiters recover: on the `UseOther` bounce, the maker's error is the `UseOther` sentinel. Propagating it through `SingletonError::source()` means each waiter's `Negotiate` future matches it via `UseOther::is()` and bounces to its own fallback leg. So sharing the error re-arms the existing fallback path for the waiters; no separate re-drive is needed.

This covers the case where the maker's future resolves `Err` but I don't think it covers the case where the maker's future is dropped before resolving: `DitchGuard::drop` still resets state to `Empty` and drops the waiters' senders, so those waiters resolve `Canceled`. That's the [`cancel_driver_cancels_all`](https://github.com/hyperium/hyper-util/blob/v0.1.20/src/client/pool/singleton.rs#L467) case with the `// TODO: cooperative baton refactor` comment.

Overall looks good, just one question. Is leaving the maker-dropped path on `Canceled` intended for this PR or do you have an idea of what that path should look like?

seanmonstar:
Correct, the other case needs the baton passing refactor. I can see about working on that this week.

PR #298, opened 2026-06-29
fix(client-pool): wake cache waiters in FIFO order

aajtodd:
## Problem

When a checkout misses the idle set, `Cache` parks the request as a waiter and returns a connection to it once one becomes available (`Shared::put`). Waiters were stored in a `Vec` and woken with `pop()` — last-in, first-out.

Under sustained load where the number of in-flight checkouts exceeds the number of reusable connections (e.g. a bounded connection pool), LIFO wakeup starves the oldest waiters: each returned connection goes to the most recently parked checkout, so a waiter that arrived early is repeatedly skipped in favor of newer ones. It is served only when the backlog drains. The result is unbounded tail latency — a small fraction of requests wait for the entire duration of the load while the median stays low.

## Fix

Store waiters in a `VecDeque` and wake from the front (`pop_front`), pushing new waiters to the back. A returned connection is handed to the longest-waiting checkout. Service order is now arrival order, so a waiter's worst-case wait is bounded by the number of waiters ahead of it rather than by how long the load lasts.

The change is local to `Shared`; the public API is unchanged.

## Effect

Measured with a load generator against a 2-node server fleet (H1, 8 MiB responses, 512 concurrent workers, 20s sustained), with the connection pool capped below the offered concurrency:

| max connections | P999 before (LIFO) | P999 after (FIFO) |
|---|---|---|
| 64 | 23.2s | 0.38s |
| 128 | 23.2s | 0.40s |
| 256 | 23.2s | 0.57s |
| 512 | 8.7s | 1.43s |

P50 stayed in the tens-of-ms range in both cases. Throughput (line-rate ~196 Gb/s on 8 MiB objects) was unchanged — the fix only affects which waiter is served when a connection returns, not how fast connections complete.

The P999 values before the fix pin at ~23s (the measurement window length), confirming that 0.1% of requests were never served a connection for the entire run — they starved. After the fix, every waiter advances in arrival order and P999 drops into the sub-second range.

## Test

`test_waiters_woken_in_fifo_order` parks three checkouts in order behind a held connection, then returns the connection one at a time and asserts they are woken oldest-first. It fails against the previous `Vec::pop` behavior (`oldest waiter was not woken first`) and passes with `VecDeque::pop_front`.

### DIFF
--- a/src/client/pool/cache.rs
+++ b/src/client/pool/cache.rs
@@ -18,6 +18,7 @@ pub use self::internal::Cached;
 // more public, but we can't change type shapes (generics) once things are
 // public.
 mod internal {
+    use std::collections::VecDeque;
     use std::fmt;
     use std::future::Future;
     use std::pin::Pin;
@@ -108,7 +109,7 @@ mod internal {
     #[derive(Debug)]
     pub struct Shared<S> {
         services: Vec<S>,
-        waiters: Vec<oneshot::Sender<S>>,
+        waiters: VecDeque<oneshot::Sender<S>>,
     }
 
     // impl Builder
@@ -149,7 +150,7 @@ mod internal {
                 events: self.events,
                 shared: Arc::new(Mutex::new(Shared {
                     services: Vec::new(),
-                    waiters: Vec::new(),
+                    waiters: VecDeque::new(),
                 })),
             }
         }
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Oh yes, good catch! This definitely should be a FIFO queue.

PR #299, opened 2026-06-30
fix(pool): Singleton shared calls pass baton if one cancels

seanmonstar:
Previously, with a bunch of calls all funneled into a `Singleton`, the first call was the "driver", and the rest all waited for the driver to finish and send a clone to them. The downside is that if the first call canceled, all the subsequent waiters would just get a "canceled" error.

This changes to a baton-passing strategy. If the first future gets canceled, it passes the baton to keep driving the future to one of the other waiters. Only once all of them are canceled will it cancel the connection attempt.

cc @aajtodd

### DIFF
--- a/src/client/pool/singleton.rs
+++ b/src/client/pool/singleton.rs
@@ -23,10 +23,9 @@
 use std::sync::{Arc, Mutex};
 use std::task::{self, Poll};
 
-use tokio::sync::oneshot;
 use tower_service::Service;
 
-use self::internal::{DitchGuard, SingletonError, SingletonFuture, State};
+use self::internal::{SingletonError, SingletonFuture, State};
 
 type BoxError = Box<dyn std::error::Error + Send + Sync>;
 
@@ -44,7 +43,7 @@ where
     M: Service<Dst>,
 {
     mk_svc: M,
-    state: Arc<Mutex<State<M::Response>>>,
+    state: Arc<Mutex<State<M::Future, M::Response>>>,
 }
 
 impl<M, Target> Singleton<M, Target>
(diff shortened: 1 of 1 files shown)


aajtodd:
Pulled the branch and ran it against the maker-future-dropped case (driver's future dropped mid-make): the surviving waiter inherits the drive and connects, where master resolves Canceled. Also confirmed the existing UseOther-bounce cancellation from #4119 stays fixed. LGTM.

One additive coverage note: the existing cancellation tests promote a waiter that was registered but not yet polled (its stored waker is None). The path where an already-polled, parked waiter is promoted on driver drop, where the promotion carries a stored Waker, isn't exercised (I think).

```rs
  #[tokio::test]
  async fn cancel_driver_promotes_parked_waiter() {
      let (mock_svc, mut handle) = tower_test::mock::pair::<(), &'static str>();
      let mut singleton = Singleton::new(mock_svc);

      std::future::poll_fn(|cx| singleton.poll_ready(cx)).await.unwrap();

      let mut fut1 = singleton.call(()); // driver
      let mut fut2 = singleton.call(()); // waiter

      // Poll both once so the waiter parks with a stored waker (not just registered).
      std::future::poll_fn(|cx| {
          assert!(Pin::new(&mut fut1).poll(cx).is_pending());
          assert!(Pin::new(&mut fut2).poll(cx).is_pending());
          Poll::Ready(())
      })
      .await;

      drop(fut1); // promote the parked waiter to driver

      let ((), send_response) = handle.next_request().await.unwrap();
      send_response.send_response("svc");
      fut2.await.unwrap();
  }
```

PR #300, opened 2026-07-02
fix(client): don't reject CONNECT tunnel responses split across reads

chatman-media:
The tunnel handshake in `client::legacy::connect::proxy::Tunnel` matches the proxy's status line against whatever bytes came back from the last `read()`, so if "HTTP/1.1 200 OK" happens to arrive split across two TCP segments, the first partial read (say, just "HTTP/1.1 2") doesn't match any of the expected prefixes and the whole thing bails out with `TunnelUnsuccessful` — even though a valid response is still on its way. Loopback tests never catch this since a single `write_all` almost always shows up as one `read`, but real proxies over a real network can and do split responses like this.

Fix just checks whether what we've got so far could still turn into a recognized status line before giving up, instead of only checking for a full/exact prefix match. Added a test that writes the response in two separate flushed writes with a delay in between (fails without the fix, passes with it), plus a test that a genuinely bad response split the same way still gets rejected promptly rather than hanging.

### DIFF
--- a/src/client/legacy/connect/proxy/tunnel.rs
+++ b/src/client/legacy/connect/proxy/tunnel.rs
@@ -224,10 +224,31 @@ where
         // else read more
         } else if recvd.starts_with(b"HTTP/1.1 407") {
             return Err(TunnelError::ProxyAuthRequired);
+        } else if is_prefix_of_expected_status_line(recvd) {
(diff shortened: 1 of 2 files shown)


cratelyn:
i believe this was superseded by #308. if this is mistaken, let me know!

PR #301, opened 2026-07-15
fix(proxy): preserve partial SOCKS response

JSap0914:
## Summary
- Preserve buffered SOCKS response bytes when a reply arrives across multiple reads.
- Advance the buffer only after a complete message has been parsed.
- Related issue: hyperium/hyper#4127

## Verification
- Focused proxy test: `cargo test --test proxy test_socks_v5_without_auth_works`
- Proxy test suite: `cargo test --test proxy`

### DIFF
--- a/src/client/legacy/connect/proxy/socks/mod.rs
+++ b/src/client/legacy/connect/proxy/socks/mod.rs
@@ -9,7 +9,7 @@ use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-use bytes::BytesMut;
+use bytes::{Buf, BytesMut};
 
 use hyper::rt::Read;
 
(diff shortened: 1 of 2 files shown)


src/client/legacy/connect/proxy/socks/mod.rs
```diff
@@ -52,7 +52,9 @@ where
         let n = crate::rt::read(&mut conn, &mut tmp).await?;
         buf.extend_from_slice(&tmp[..n]);
 
-        match M::try_from(buf) {
+        let mut message = buf.clone();
```
seanmonstar, line 55: This will result in copying all the bytes. It would be better to check in the parse function if remaining has enough bytes to parse.

JPDye:
If this isn't actively being worked on, I'll take over and fix it, since it's my mistake in the first place.

JPDye:
I've put up a branch that solves this. (sorry if I stepped on some toes, but I assumed after 2 weeks it's probably alright).

#308

PR #302, opened 2026-07-22
Handle IPv6 addresses as SOCKS destination

FrankTaylorLieder:
(Note: I could not find docs on how to contribute to this repo, let me know if there is a process to follow.)

This change fixes an issue when accessing IPv6 services through the SOCKS proxy using their IP address rather than hostname:
- The IP address is passed over to the SOCKS proxy as type ATYP=0x03 (Domain name) instead of ATYP=0x04 (Address).
- The remote end therefore tries to DNS resolve the address, which fails.

The root cause is:
- The target URI is of the form `http://[2001:db8::1]:8080`.
- The proxy extracts the host part of the URI and gets back `[2001:db8::1]`. (`src/client/legacy/connect/proxy/socks/v4/mod.rs:133`)
- It attempts to parse this as an address and fails, so assumes its a domain name. Incorrect.
- This is not a valid IPv6 address due to the square brackets (they are part of the URI encoding of IPv6 addresses). These need to be stripped to make a valid IPv6 address.

Fix:
- Strip any square brackets from the host part of the URI.
- There will never be square brackets in valid domain names or IPv4 addresses.

### DIFF
--- a/src/client/legacy/connect/proxy/socks/v4/mod.rs
+++ b/src/client/legacy/connect/proxy/socks/v4/mod.rs
@@ -130,7 +130,19 @@ where
 
         let fut = async move {
             let port = dst.port().map(|p| p.as_u16()).unwrap_or(443);
-            let host = dst.host().ok_or(SocksError::MissingHost)?.to_string();
+            let host_from_uri = dst.host().ok_or(SocksError::MissingHost)?;
+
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):
Awesome, thanks!

PR #187, opened 2025-04-25
feat(client): add proxy::SocksV4 and proxy::SocksV5 connectors

JPDye:
cc hyperium/hyper#3851, hyperium/hyper#3877 hyperium/hyper-util#173

Adds `SocksV4` and `SocksV5` connectors with simple configuration for local and remote DNS. Supports SOCKSv5 user/pass authentication and allows optimistic message sending (not in RFC but supported by most servers) to reduce number of round-trips in SOCKSv5 handshake.

TODO:
- [X] Properly test optimistic sending against real server implementations
- [X] Cleanup error handling

PR #307, opened 2026-07-30
fix(client): write a single NULL after USERID in SOCKS4 requests

monosans:
## What's wrong

The SOCKS4 connector writes one byte too many in its CONNECT request.

A SOCKS4 request is an 8-byte header followed by a **user ID** — a variable-length string ended by a single zero byte:

```
VN(1) CD(1) DSTPORT(2) DSTIP(4) USERID(variable) NULL(1)
```

— [_SOCKS: A protocol for TCP proxy across firewalls_](https://www.openssh.com/txt/socks4.protocol), Ying-Da Lee

This connector never sends a user ID, so that field is empty and only its single terminating zero byte belongs on the wire — 9 bytes in total. `Request::write_to_buf` in `src/client/legacy/connect/proxy/socks/v4/messages.rs` wrote _two_ zero bytes (one commented `USERID`, one commented `NULL`), so it sent 10.

## Why it matters

The SOCKS4a extension — the one that lets the _proxy_ resolve a domain name — puts the hostname immediately after that terminator, itself ended by a zero byte:

```
VN(1) CD(1) DSTPORT(2) DSTIP(4) USERID(variable) NULL(1) HOSTNAME(variable) NULL(1)
```

— [_SOCKS 4A: A Simple Extension to SOCKS 4 Protocol_](https://www.openssh.com/txt/socks4a.protocol), Ying-Da Lee

That is this connector's default mode: `local_dns` is off by default, so a domain destination is sent by name for the proxy to resolve. With an extra zero byte in front of it, the proxy reads the hostname field as an empty string, and the real hostname is never part of the request at all.

Feeding both versions of the bytes to a parser that follows the spec literally:

```
SOCKS4a BEFORE (19 bytes): port=443 ip=0.0.0.255 userid="" hostname=""         leftover="hyper.rs\0"
SOCKS4a AFTER  (18 bytes): port=443 ip=0.0.0.255 userid="" hostname="hyper.rs" leftover=""
SOCKS4  BEFORE (10 bytes): port=443 ip=127.0.0.1 userid="" (no hostname)       leftover="\0"
SOCKS4  AFTER  ( 9 bytes): port=443 ip=127.0.0.1 userid="" (no hostname)       leftover=""
```

In other words:

- **SOCKS4a (domain destination — the default) is broken.** The proxy gets an empty hostname and the request fails. Worse, the leftover `hyper.rs\0` stays in the proxy's receive buffer and is relayed into the tunnel as if the client had sent it as application data.
- **Plain SOCKS4 (IPv4 destination) still connects**, because the address travels in DSTIP — but the leftover zero byte is still there and reaches the target as a stray first byte of the tunneled stream.

As an independent check, curl builds the same request with exactly one terminator: in `lib/socks.c` it appends `strlen(user) + 1` bytes for the user ID (just the terminator when there is no user), then `strlen(hostname) + 1` for SOCKS4a.

## The change

- Remove the extra zero byte, and adjust the two `remaining_mut()` checks and the returned lengths from `10` to `9`.
- Add `test_socks_v4a_with_server_resolved_domain_works`. The domain path had no test at all, which is how this went unnoticed since the connectors landed in #187 — the existing `test_socks_v4_works` only covers the IPv4 path, where the bug is much less visible.
- Fix one comment on the line being touched: `buf.put_u16(*port)` was labelled `// IP`.

No public API change; the only behavioural difference is one fewer byte on the wire. `cargo test --all-features --test proxy` passes.

### DIFF
--- a/src/client/legacy/connect/proxy/socks/v4/messages.rs
+++ b/src/client/legacy/connect/proxy/socks/v4/messages.rs
@@ -9,7 +9,7 @@ use std::net::SocketAddrV4;
 /// |  1  |  1  |    2    |         4         |   Variable  |  1   |  Variable  |   1  |
 /// +-----+-----+----+----+----+----+----+----+-------------+------+------------+------+
 ///                                                                ^^^^^^^^^^^^^^^^^^^^^
-///                                                      optional: only do IP is 0.0.0.X
+///                                                      optional: only if IP is 0.0.0.X
 #[derive(Debug)]
 pub struct Request<'a>(pub &'a Address);
 
@@ -41,7 +41,7 @@ impl Request<'_> {
     pub fn write_to_buf<B: BufMut>(&self, mut buf: B) -> Result<usize, SerializeError> {
         match self.0 {
             Address::Socket(socket) => {
-                if buf.remaining_mut() < 10 {
+                if buf.remaining_mut() < 9 {
                     return Err(SerializeError::WouldOverflow);
                 }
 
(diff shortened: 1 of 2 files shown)


seanmonstar:
cc @JPDye what do you think?

JPDye:
It (the code) is correct.

seanmonstar (approved):

Smooveemaan:
Thanks for this fix! Could it go out in a 0.1.21 release?

In 0.1.20, `SocksV4` writes two NULs after the empty USERID, which breaks both SOCKS4 variants when going through reqwest:

- **SOCKS4a** (`socks4a://`): the proxy reads the domain as an empty string, so every connection by host name fails.
- **Plain SOCKS4** (`socks4://`): the proxy treats the second NUL as application data and forwards it to the destination. With 3proxy in front of Apache, the HTTP request that follows gets `400 Bad Request`, and a TLS ClientHello is corrupted the same way.

Reproduction against 3proxy (SOCKS4 CONNECT to an Apache container on port 80, then `HEAD / HTTP/1.0`):

```
[4, 1, 0, 80, ip.., 0]     -> HTTP/1.1 401 Unauthorized  (expected response)
[4, 1, 0, 80, ip.., 0, 0]  -> HTTP/1.1 400 Bad Request   (what 0.1.20 sends)
```

We're working around it for now with a loopback SOCKS5 bridge in front of our own SOCKS4 client, and would like to drop that once a release includes the fix.

monosans:
> Thanks for this fix! Could it go out in a 0.1.21 release?

While I’m also waiting for the release, I’ve added the following to Cargo.toml:

```toml
[patch.crates-io]
hyper-util = { git = "https://github.com/hyperium/hyper-util", branch = "master" }
```

PR #312, opened 2026-08-17
feat(client): add UnixConnector for Unix domain sockets

WhyNotHugo:
Connector for the legacy Client that always connects to a fixed Unix domain socket path, analogous to HttpConnector for TCP.

Useful for talking to locally running services, special proxies, or tunnels.

Destination Uri is used only for the Host header and request target.

- [x] The description, docs, and comments (words meant for humans) are written by a human, not by an LLM. (https://seanmonstar.com/micro/20260729-i-want-your-own-words/)

### DIFF
--- a/src/client/legacy/connect/mod.rs
+++ b/src/client/legacy/connect/mod.rs
@@ -4,6 +4,8 @@
 //!
 //! - A default [`HttpConnector`][] that does DNS resolution and establishes
 //!   connections over TCP.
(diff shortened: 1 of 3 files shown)


WhyNotHugo:
CI failures don't seem related to my changes; they're unused imports for pre-existing lines.

WhyNotHugo:
I don't really understand what the CI failure means; it indicates that only IPv4 and IPv6 are supported, but not Unix sockets?

PR #315, opened 2026-08-22
fix(client): parse proxy CONNECT response with httparse

aashish-thapa:
Closes #4095.

The CONNECT tunnel validated the proxy's response by pattern-matching the buffer instead of parsing it, which caused two problems:

- **Malformed responses were accepted.** `starts_with(b"HTTP/1.1 200")` only checks a prefix, so a status like `HTTP/1.1 2000 OK` was treated as a successful tunnel.
- **Server-speaks-first responses broke.** Success required the whole buffer to end with `\r\n\r\n`. When the destination's first bytes arrive in the same read (`HTTP/1.1 200 OK\r\n\r\n<data>`), that check fails, so the loop keeps reading for a response that already came and stalls until EOF.

This parses the response with `httparse` (the parser hyper already uses). The status line is validated properly, and `Status::Complete` locates the end of the headers even when data trails them, so the tunnel establishes correctly.

One thing left out on purpose: the early bytes are now recognized but not yet handed back to the caller. Returning them means replaying the buffered data through something like `Rewind`, which changes the public `Service::Response` type. Happy to do that as a follow-up if you want it, but it felt like a separate, breaking decision rather than part of this fix.

The first commit adds tests that fail against the current heuristic; the second applies the fix and they pass.


aashish-thapa:
@seanmonstar  would you mind reviewing it when you get a chance please.

seanmonstar (approved):
Thanks!

PR #320, opened 2026-09-11
feat(client): add HttpConnect connector that keeps early tunnel data

### LINKED ISSUE #4095: HTTP CONNECT proxy tunnel has lax response parsing and loses early data
### Version

hyper-util 0.1.20

### Platform

Linux 6.18.14

### Summary

While evaluating `hyper-util`'s [client::legacy::connect::proxy::tunnel](https://github.com/hyperium/hyper-util/blob/911d0f256342fa6740721104418c6102dc7d1d0b/src/client/legacy/connect/proxy/tunnel.rs#L167) for use with gRPC routing over HTTP CONNECT proxies, I identified two limitations in the current handshake implementation.

## 1. Lax Response Parsing
It [heuristically matches](https://github.com/hyperium/hyper-util/blob/911d0f256342fa6740721104418c6102dc7d1d0b/src/client/legacy/connect/proxy/tunnel.rs#L217-L218) the start and end of the expected response, potentially accepting malformed responses.

This can be fixed by using `httparse` to properly parse the response.

## 2. No Buffering for Early Data
It assumes the read buffer contains no data from the target server immediately after the proxy's response. This assumption holds for protocols where the client always speaks first (e.g., TLS) after the tunnel is established. However, since gRPC implementations must work with any security protocol, they must handle cases where the server sends data immediately.

Fixing this would require an API change, as the returned I/O stream would need to incorporate a buffer for the peeked data, similar to the private [Rewind](https://github.com/hyperium/hyper-util/blob/911d0f256342fa6740721104418c6102dc7d1d0b/src/common/rewind.rs#L13) struct.

Although the handshaking code is simple enough to be re-implemented in gRPC, I wanted to check if these fixes can be upstreamed to allow us to use `hyper-util`.

### Code Sample

```rust

```

### Expected Behavior

* The proxy response must be correctly parsed.
*  The data from the target server must be retained.

### Actual Behavior

* A malformed proxy response may be accepted
* The data from the target server may be lost.

### Additional Context

_No response_

heitzlki:
Follow-up to hyperium/hyper#4095: #315 fixed the parsing, but `tunnel()` still throws away anything after the header end, which breaks protocols where the server speaks first. Fixing it in place changes `Tunnel`'s response type, which is why #313 went nowhere.

So this is the alternative @seanmonstar suggested on the issue: a new connector, `HttpConnect`, that sends an actual CONNECT request over an h1 connection and returns the upgraded IO, early bytes survive through hyper's own upgrade buffer. The conn future is driven inside the connector future, so nothing needs to be spawned. The returned IO forwards `Connected` so it works with the legacy client. `Tunnel` itself is untouched except a `pub(super)` on the shared `Headers` enum.

HTTP/1 only for now, h2 CONNECT could go behind the same type later.

Closes hyperium/hyper#4095


src/client/legacy/connect/proxy/http_connect.rs
```diff
@@ -0,0 +1,644 @@
...
+
+        self
+    }
+}
+
+impl<C> Service<Uri> for HttpConnect<C>
+where
+    C: Service<Uri>,
+    C::Future: Send + 'static,
+    C::Response: Read + Write + Connection + Unpin + Send + 'static,
```
seanmonstar, line 143: So, here's an idea. (It might have been in my head and I never expressed it.) This `HttpConnect` type could be generic over an HTTP service, instead of a "connector", and then it could work with HTTP/1 or 2 (or 3?).

That'd mean `impl Service<http::Request<SomeEmptyBody>, Response = http::Response<B>>`, and then we would build a `CONNECT` request kind of like in the handshake function, using `hyper::upgrade`. Of course, the user would have to provide an HTTP connection that supports upgrades (so if it's HTTP/2, they'd need to have enabled extended connect). But, that seems fine.

Is this too crazy?

heitzlki: Not too crazy, h2 CONNECT multiplexing tunnels over one proxy connection is exactly what the gRPC folks want anyway. Two questions before I rework it:

1 Should the service-generic type be the whole API, or the core with the current connector version kept as a thin wrapper? With a bare service the caller has to drive the conn themselves, which is the one thing the connector version handles for you.

2 Behind a service I can't see the transport, so Tunneled loses the underlying Connected. Fine to just return Connected::new()?

PR #321, opened 2026-09-12
feat(client): Add Expect: 100-continue body wrapper

### LINKED ISSUE #3833: Update Hyper Client to support handling `expect: 100-continue` header
**Is your feature request related to a problem? Please describe.**

When I use Hyper Client based on the example (https://github.com/hyperium/hyper/blob/3817a79b213f840302d7e27fac8508de9caada0f/examples/client.rs#L49-L54), it doesn't correctly handle the `expect: 100-continue` header in the request. 

**Describe the solution you'd like**

Hyper Client needs to wait for the `100 continue` response from the S3WSServer after sending the request header parts for certain time and then send the request body if it gets the response from the server or after certain time threshold.

This is to have the same functionality as https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/100

**Describe alternatives you've considered**

Expose the client parts where it sends the request header and body, but it will be inefficient for the developers to implement the same functionality.

**Additional context**

bountis:
Add a request body wrapper which handles the `Expect: 100-continue` header flow.
https://www.rfc-editor.org/rfc/rfc9110.html#name-expect

Also adds a utility `wrap` which wraps a request body in `ExpectContinueBody`, adds the `Expect: 100-continue` header if not present, adds an `on_informational` hook which signals to release the body.
The `poll_frame` implementation parks the body until a signal is received/cancelled or a configured timeout elapses.

Tests and a small example are added which demonstrate the behavior.

Closes https://github.com/hyperium/hyper/issues/3833


src/client/expect_continue.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+/// wrap returns a request whose body is withheld until 100/timeout
+///
+/// It adds `Expect: 100-continue` header to the request if absent.
+/// Wraps the body in `ExpectContinueBody`.
+/// Registers an `on_informational` hook that releases the body on a 100.
+pub fn wrap<B>(
+    req: Request<B>,
+    sleep: Option<Pin<Box<dyn Sleep>>>,
```
seanmonstar, line 97: Why accept an `Option`? Should we not nudge people harder to supply a timeout?

bountis: Hmm, yes, this should probably be stricter and require a timeout as the RFC also states that `client SHOULD NOT wait for an indefinite period before sending the content`.

I can modify `wrap` then to accept a timer + a duration. `ExpectContinueBody::new` can probably continue to accept the option since it's private to the crate and the signal/cancel tests use `None`, or do you want me to change that too?

